PR #68, opened 2023-03-20
Fix graphql schema to support different zkApp transactions within a block

MartinMinkov:
## Description

Since there can be multiple zkApp transactions within a block, assuming that events/actions will be under the duplicate zkApp transactions within a block is always wrong. We change the schema here to support different zkApp transactions with event/action data within the same block.


bkase (approved):
Looks good 👍

PR #71, opened 2023-04-03
Add actions filtering based on `fromActionHash` and `endActionHash` + return all action states for each block

MartinMinkov:
## Description
:link: Laying the groundwork for o1-labs/snarkyjs#802

In this PR, we introduce an enhancement that enables users to filter actions based on their action hash within each block. By specifying fromActionHash or endActionHash, users can now filter actions according to the action hash. This improvement allows SnarkyJS users to simply provide the desired action hashes, eliminating the need to derive them from the entire actions history. As a result, SnarkyJS experiences a performance boost, since it no longer needs to compute a large number of hashes. 

Additionally, we return all action states for each block. SnarkyJS can use `actionStateTwo` to recalculate the new actions hash.


schema.graphql
```diff
@@ -12,6 +12,16 @@ input EventFilterOptionsInput {
   from: Int
 }
 
+input ActionFilterOptionsInput {
+  address: String!
+  tokenId: String
+  status: BlockStatusFilter
+  to: Int
+  from: Int
+  fromActionHash: String
```
mitschabaude, line 21: could we rename this to `fromActionState` / `endActionState` for consistency with the rest of the protocol? I'm going to argue for the same on the SnarkyJS side

mitschabaude (approved):

PR #80, opened 2023-04-24
Improved Filtering and Selection Logic for Best Tip and Tie Breaker Mechanism

MartinMinkov:
# Description
In this PR, we have enhanced the filtering and selection logic for the best tip and refactored the code for better readability and maintainability. The primary focus is on handling multiple blocks at the highest tip of the network (the latest block height) and implementing a tiebreaker mechanism to decide the optimal block to proceed with. This is crucial in the context of events/actions, as users rely on the data emitted in the latest block available, requiring an accurate representation of the network.

The tie-breaking process for blocks with duplicate heights at the maximum level is discussed in detail here:
[Consensus Tie Breaker Rules](https://github.com/MinaProtocol/mina/blob/36d39cd0b2e3ba6c5e687770a5c683984ca587fc/docs/specs/consensus/README.md?plain=1#L1134)

As outlined, the `lastVRF` property for each block is examined, and the highest Blake2B hash value is selected. In the unlikely event of a tie, the higher block hash is chosen. The select function in the Mina daemon illustrates these rules:

[Select based on VRF Output](https://github.com/MinaProtocol/mina/blob/bff1e117ae4740fa51d12b32736c6c63d7909bd1/src/lib/consensus/proof_of_stake.ml#L3004)
[Select based on Block Hash](https://github.com/MinaProtocol/mina/blob/bff1e117ae4740fa51d12b32736c6c63d7909bd1/src/lib/consensus/proof_of_stake.ml#L3001)
To facilitate this process, we introduced the blakejs dependency, which is used to construct a Blake2B hash and convert it into hex format for each lastVRFOutput. We then analyze all the latest blocks and run the comparison function.


bkase (changes requested):
This change looks good to me but we should definitely ask @nholland94 to confirm.

However, before we land this we need to do something to mitigate inevitable incompatibilities that will creep in as the consensus rules change.

Given that we don’t want to introduce a wasm nor jsofocaml runtime dependencies to this server, what I think we need is some kind of diff test between another well-tested implementation and this one. The choices are: the real implementation in Mina OR the selection algorithm in openmina (after verifying that that implementation is also bound to the main implementation in some way). This is one of those times I wish we had some kind of executable spec that we could derive multiple implementations from.

deepthiskumar:
Sounds like we should start planning for a framework to testing multiple implementations?
In the short-term for this PR we could do something where the select function is exposed in an app on both ocaml and ts impl that takes a set of block files (json format block files mina nodes export) and returns the selected state hash. The test script could then call each of the apps with the same blocks and then compare the output?

PR #94, opened 2023-11-21
feat(docker-compose.dev.yml): make dev workflow easier

MartinMinkov:


docker-compose.dev.yml
```diff
@@ -1,6 +1,28 @@
 version: '3.3'
 
 services:
+  mina:
+    image: ${MINA}
+    container_name: mina
+    ports:
+      - '6060:6060'
+      - '8302:8302'
+      - '127.0.0.1:3085:3085'
```
Trivo25, line 10: is `127.0.0.1:3085:3085` correct or should this be `3085:3085`?

Trivo25 (approved):

docker-compose.dev.yml
MartinMinkov: It does work; the difference between the two is that with `127.0.0.1:3085:3085`, it specifies that the port is only bound to localhost, and will only be accessible by the host machine, while `3085:3085` is accessible by any machine on the network. This detail doesn't matter in this context, but I changed it to `3085:3085` since I think it's more clear :+1:

PR #99, opened 2024-01-30
Add lightnet to tests

MartinMinkov:


.github/workflows/run-tests.yaml
```diff
@@ -9,42 +9,30 @@ on:
 jobs:
   Run-Tests:
     runs-on: ubuntu-latest
-    container: node:latest
     services:
-      postgres:
-        image: postgres:latest
+      mina-local-network:
```
shimkiv, line 13: We might also want to attach [logs volume](https://github.com/o1-labs/o1js/blob/baac0cbd8e0e37d51c1db1acc6d607ca96af4750/.github/workflows/live-tests.yml#L32) so that after the tests we can upload files to [CI run](https://github.com/o1-labs/o1js/blob/baac0cbd8e0e37d51c1db1acc6d607ca96af4750/.github/actions/live-tests-shared/action.yml#L31). This will improve debugging since Mina related logs will be available for download right away.

shimkiv (approved):
Great stuff, have one comment about CI config.

PR #108, opened 2024-10-22
Upgrade o1js version

45930:
Upgrade o1js to modern version and make necessary changes for tests to run on lightnet


.github/workflows/run-tests.yaml
```diff
@@ -11,7 +11,7 @@ jobs:
     runs-on: ubuntu-latest
     services:
       mina-local-network:
-        image: o1labs/mina-local-network:o1js-main-latest-lightnet
+        image: o1labs/mina-local-network:compatible-latest-lightnet
```
45930, line 14: `o1labs/mina-local-network:o1js-main-latest-lightnet` is not a valid image name, see: https://github.com/o1-labs/Archive-Node-API/actions/runs/11448838391/job/31853195062

zkapp/utils.ts
```diff
@@ -69,6 +70,9 @@ async function deployContract(
   let transaction = await Mina.transaction(
     { sender, fee: transactionFee },
     async () => {
+      if (fundNewAccount) {
+        AccountUpdate.fundNewAccount(sender);
```
45930, line 74: This line is required in general, but must be omitted if we deploy to a keypair that already exists, as is the case in `tests/resolvers.test.ts#102`

deepthiskumar (approved):

PR #110, opened 2024-11-21
Updating GraphQL resolver tests

45930:
Another one to reduce the complexity of reviewing https://github.com/o1-labs/Archive-Node-API/pull/107

This PR:

- exposes port 8282 on docker in CI, which is used by lightnet
- improves the resolver tests
  - properly annotates the type of the graphQL response object
  - exits the process with code `1` on error instead of always exiting with code `0`
  - makes use of additional describe blocks to clarify the test cases
  - explicitly test response data for correctness, rather than metadata like length only

Since there are no implementation changes here, we can be more confident about changes to the test file, then in the other PR where I do change the implementation, I will be able to leave existing tests as-is.

### DIFF
--- a/tests/resolvers.test.ts
+++ b/tests/resolvers.test.ts
@@ -3,7 +3,11 @@ import assert from 'node:assert';
 import { createYoga, createSchema } from 'graphql-yoga';
 import { loadSchemaSync } from '@graphql-tools/load';
 import { GraphQLFileLoader } from '@graphql-tools/graphql-file-loader';
-import { buildHTTPExecutor } from '@graphql-tools/executor-http';
+import {
+  buildHTTPExecutor,
+  HTTPExecutorOptions,
+} from '@graphql-tools/executor-http';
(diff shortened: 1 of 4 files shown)


tests/resolvers.test.ts
```diff
@@ -79,179 +112,263 @@ query getActions($input: ActionFilterOptionsInput!) {
...
+        eventsResponse = results.data.events;
+        lastBlockEvents = eventsResponse[eventsResponse.length - 1].eventData!;
+      });
+      test('GQL response contains multiple events in the latest block', async () => {
+        assert.strictEqual(lastBlockEvents.length, numberOfEmits);
+      });
+      test('the events have the correct data', async () => {
+        for (let i = 0; i < numberOfEmits; i++) {
+          const eventData = lastBlockEvents[i]!;
+          // The event type is 1 and the event data is 2, 1 (Bool(true)), and the zkapp address
```
deepthiskumar, line 289: ... 1, and the zkapp address

deepthiskumar (approved):

PR #115, opened 2025-02-28
Filter pending blocks as well as canonical

45930:
This PR applies the block height filter to the pending blocks as well as canonical.

Currently, if a user filters for actions from blocks a to b, they will get those _plus_ all actions that are in pending blocks, regardless of their block height.

We should filter the canonical and the pending chains so that users who make a specific request get only the data they asked for.  In the default case, there is no difference since the last `n` blocks will aways included the pending chain.

OLD:
image

NEW: 
image
image


45930:
Re: changelog, we never actually release new versions of this... staketab just grab the commit and run it from source.  I think it's a good idea to tag versions and keep a changelog.  We will have to keep ourselves honest.

45930:
Merging this now to solve a need for zknoid

PR #116, opened 2025-03-07
move pending blocks query out of recursive CTE

45930:
Follow on to #115 .....

This PR moves the query our of the recursive CTE and into the full chain union query.  The height referenced in the recursive CTE is not usable as I was trying previously.

Example screenshots from the old code:

image

image

New code:
image


boray (approved):
Note: We need to add tests to catch edge cases as discussed.

PR #117, opened 2025-04-14
Archive node devex improvements

hattyhattington17:
- Use mainnet/devnet prefix for DB dumps
- Make docker-compose DB init script executable
- Update dump downloader to support mainnet/devnet and fallback to previous day's dump if today's is missing 
    - must update `PG_DUMP` env var to use devnet


scripts/init_docker_compose.sh
hattyhattington17, line 1: I only changed the permissions on this file! Github seems to have a bug https://github.com/orgs/community/discussions/6814

scripts/download_db.sh
```diff
@@ -1,26 +1,52 @@
...
-# Change to the data directory
-cd ${DATA_DIR}
+# get date as YYYY-MM-DD
+get_date() {
+  # macOS
+  if [[ "$OSTYPE" == "darwin"* ]]; then
+    date -v -"$1"d '+%Y-%m-%d'    
+  # linux
+  else
+    date -d "-$1 days" '+%Y-%m-%d'
```
45930, line 29: The mac thread works on my machine TM.  Trusting that the linux branch also works.

45930 (approved):

PR #125, opened 2025-10-15
suggestions for additional improvements

duccas:


scripts/generate_libp2p.sh
```diff
@@ -9,26 +9,51 @@
...
 
 # Enable debug mode
 set -x
 
 # Exit on error, and pipefail
 set -eo pipefail
 
+# Get script directory and load .env from project root
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+PROJECT_ROOT="$(dirname "$SCRIPT_DIR")"
```
glyh, line 24: Suggest change:
```bash
PROJECT_ROOT="$(git rev-parse --show-toplevel)"
```
So if we move the scripts dir around it stil works.

scripts/generate_libp2p.sh
```diff
@@ -9,22 +9,36 @@
...
+  echo "  docker - Generate keypair using Docker only"
+  echo "  binary - Generate libp2p keypair using mina binary only"
+  exit 1
+fi
 
 # Set environment variable
 # Make sure these match the environment variables used in the docker-compose.yml file
-export MINA_PRIVKEY_PASS="passlib"
-export MINA_LIBP2P_PASS="passlib"
+export MINA_LIBP2P_PASS=$MINA_LIBP2P_PASS
```
glyh, line 46: These are 2 different variables and could be different. I'd suggest also add some panics if either is missing from env vars.

scripts/generate_libp2p.sh
```diff
@@ -9,22 +9,36 @@
...
 
 # Constants
 KEYPAIR_DIR="keys"
 KEYPAIR_NAME="libp2p-keys"
-MINA_KEYPAIR_IMAGE="minaprotocol/mina-generate-keypair:1.3.0-9b0369c"
+MINA_KEYPAIR_IMAGE=$MINA
```
glyh, line 52: Nit: we no longer need this package, I think. mina's main docker has this functionality bundled.

scripts/download_db.sh
```diff
@@ -16,8 +16,22 @@ DB_DIR="db"
 PG_DUMP="archive.sql"
 BASE_URL="https://storage.googleapis.com/mina-archive-dumps"
 
+# Detect which docker compose command to use
+get_docker_compose_cmd() {
```
glyh, line 20: old docker compose is deprecated, I'd just replace this with `docker compose`

dkijania (approved):

SanabriaRusso:
Thanks for these @duccas — the dev-setup cleanups are appreciated, and none of this touches the GraphQL API surface (schema/resolvers/CORS), so no server-side compatibility concerns. One heads-up: the branch has drifted behind `main` since October, and #158 ("refresh stale Mina references — point to mainnet 3.3.1", `28e0cf8`) has since landed and overlaps two of the three files here:

- **`.env.example.compose`** — `main` is already on `minaprotocol/mina-daemon:3.3.1-7b34378-bullseye-mainnet` with the mainnet `PEER_LIST_URL`. This PR would roll that back to `3.2.0-…-focal`, so this file is best dropped (it's the source of the current conflict).
- **`scripts/generate_libp2p.sh`** — #158 already reworked this to generate both keypairs from the mina-daemon image (docker-only, no local `mina` binary needed), which covers the retired-`mina-generate-keypair` fix this PR was after.

The piece still worth landing is the **`download_db.sh`** change: `main` still hardcodes `docker-compose stop postgres` (Compose v1, EOL since 2023), while our own `docs/getting-started.md` tells users to run `docker compose up` (v2) — so on a v2-only host the docs work but this script fails with "command not found". Your `get_docker_compose_cmd()` detection fixes exactly that.

Suggestion: rather than rebasing the whole branch, would you be up for a small fresh PR with just the `download_db.sh` detection hunk? That lands the real improvement cleanly and we can close this one out. Happy to help either way.

dkijania:
Thanks again @duccas for these — following up on @SanabriaRusso's review.

The `download_db.sh` Compose-v2 detection was the real fix here, and I've landed it on a clean branch off `main` in **#202**, crediting this PR as its source. The other two files have since been overtaken on `main`:

- **`.env.example.compose`** — #158 moved it to `minaprotocol/mina-daemon:3.3.1-…-bullseye-mainnet`; this PR would roll it back to `3.2.0-…-focal` (the source of the current conflict).
- **`scripts/generate_libp2p.sh`** — #158 reworked it to generate both keypairs from the mina-daemon image (docker-only), covering the retired-`mina-generate-keypair` fix.

So with the `download_db.sh` improvement now carried forward in #202, I'd suggest **closing this one out** to clear the conflict. Really appreciate the dev-setup cleanups — the Compose-v2 detection in particular fixes a real break for anyone on a v2-only host following our own docs.

SanabriaRusso:

Blocking for a mechanical reason, not a quality one: the branch is `CONFLICTING` / `DIRTY` against `main` (`e353bbc`), and two of its three files have since been overtaken on `main` such that merging would be a **rollback**. The one genuinely useful change here is already carried forward in an open, clean PR. Recommendation is unchanged from @dkijania's last comment: **close this**, land #202.

### What I checked

- **Downstream compatibility: not applicable — this PR has zero GraphQL surface.** Files touched are `.env.example.compose`, `scripts/download_db.sh`, `scripts/generate_libp2p.sh`. No `schema.graphql`, no resolver, no `src/db/sql/events-actions/*`, no `src/services/*`, no error strings, no CORS/`maskedErrors` config. So it falls in *neither* bucket: it can't affect `mina-explorer`'s `blocks(...)`/`networkState` path, and it can't affect the o1js `events`/`actions` path. Nothing to verify against either consumer.
- **Conflict surface vs in-flight work: none.** #209 and #134 both touch `src/db/sql/events-actions/queries.ts`; this PR touches no file either of them does. Ordering against the P0–P2 series is irrelevant here — the conflict is purely with what already landed.
- **Head is unchanged since the approval, but the base moved.** Head is still `10c1a8c` (2025-10-15). @dkijania's approval is on that exact commit, so it's not stale w.r.t. the head — but it was submitted 2026-03-31, and **#158 (`chore: refresh stale Mina references — point to mainnet 3.3.1`) merged 2026-05-07**, after the approval. The approval was valid against a `main` that no longer exists.
- **CI is red but uninformative.** `Check Schema` and `build-and-deploy` both FAILURE, from workflow runs on 2025-10-15; logs are expired (GitHub returns HTTP 410), so they can't be triaged. Moot while the branch can't merge.

### The three files, individually

**1. `.env.example.compose` — regression, do not merge.**
`main` is on `minaprotocol/mina-daemon:3.3.1-7b34378-bullseye-mainnet` / `minaprotocol/mina-archive:3.3.1-7b34378-bullseye-mainnet` with the mainnet `PEER_LIST_URL`. This PR sets `3.2.0-97ad487-focal-mainnet`. That's a downgrade, and it's the source of the merge conflict. Correct in Oct 2025, wrong now.

**2. `scripts/generate_libp2p.sh` — superseded, and the head version has real gaps.**
#158 rewrote this to generate **both** keypairs from the mina-daemon image, docker-only. The version here would replace that with something narrower and, in places, weaker:
- It only generates *one* key per run and requires a mandatory `docker|binary` mode arg; `main`'s version generates the Mina account keypair *and* the libp2p keypair unconditionally. The `binary` mode reintroduces a dependency on a locally-installed `mina` binary, which #158 deliberately removed.
- `export MINA_LIBP2P_PASS=$MINA_LIBP2P_PASS` is a self-assignment. If `.env` doesn't define it, this silently exports an empty string and you get a libp2p key with an **empty passphrase** — `set -eo pipefail` won't catch it. `main` hardcodes `passlib` to match `docker-compose.yml`, which is at least deterministic.
- The docker invocation is `$MINA libp2p generate-keypair`, whereas `main` (and the Mina docs) use `advanced generate-libp2p-keypair`. I couldn't verify the `libp2p generate-keypair` form exists on `3.3.1`; if it doesn't, this fails at runtime.
- `set -x` is retained while now sourcing `.env`, so `MINA_LIBP2P_PASS` / `PGPASSWORD` get echoed to stdout. Local-dev only, but pointless.

Recommend dropping this file entirely.

**3. `scripts/download_db.sh` — this is the real fix, and it's still unlanded.**
Confirmed the bug is live on `main` today: `scripts/download_db.sh:20` is `docker-compose stop postgres` (Compose v1, EOL June 2023), while `docs/getting-started.md:141` and `:157` tell users to run `docker compose up` (v2), and `docs/getting-started.md:132` tells them to run this script. On a v2-only host the documented flow breaks with `docker-compose: command not found`. The `get_docker_compose_cmd()` detection here fixes exactly that.

That hunk was extracted to **#202** (`fix(scripts): detect docker compose v2 in download_db.sh`, head `a1052e5`), which is **still OPEN**, `mergeable: MERGEABLE`, `mergeStateStatus: BEHIND` — i.e. it just needs an update-branch, then it merges clean. Nothing is lost by closing this PR **provided #202 actually lands**; right now neither is merged, so the bug is still shipping.

### Action

1. Merge #202 (update-branch first — it's only BEHIND, no conflicts). That preserves @duccas's contribution.
2. Close #125 once #202 is in.

No fix patch is offered for this branch because the correct end state is not a rebase — after dropping files 1 and 2 the diff is exactly #202, which already exists.

### Non-blocking nits

- If #202 is ever reworked, `command -v docker &> /dev/null && docker compose version &> /dev/null` runs `docker compose version` on every invocation, which is a daemon round-trip; `docker compose version --short` is marginally cheaper. Cosmetic.
- Neither script has test coverage and none is warranted — they're local dev bootstrap, not shipped code paths.

glyh (commented):
I just realized there's pending reviews I haven't subbmited from long time ago. not sure if they're complete or still relevant.

SanabriaRusso:
Closing this in favour of #202, per the second-pass review above.

To be clear about why, since this PR was approved and the approval is still standing: the head here never drifted — the base did. #158 landed on 2026-05-07 and overtook two of the three files, so merging this now would roll `.env.example.compose` back from mina-daemon `3.3.1-bullseye-mainnet` to `3.2.0-focal`. That regression is the blocker, not anything wrong with the contribution itself.

The genuinely useful change here — Compose v2 detection in `scripts/download_db.sh` — is already open as a clean, conflict-free PR in #202, which is approved and ready once it is brought up to date. Nothing is lost by closing this.

Thanks @duccas — the underlying bug you spotted is real and still live on `main` (`scripts/download_db.sh:20` invokes v1 `docker-compose` while `docs/getting-started.md:141,157` instruct v2, so the documented flow breaks on v2-only hosts). #202 fixes exactly that.

Two things from the review worth carrying forward separately if anyone picks up `scripts/generate_libp2p.sh` again, since they are still unfixed on `main`:

- `export MINA_LIBP2P_PASS=$MINA_LIBP2P_PASS` is a self-assignment that silently yields an **empty passphrase** when `.env` omits the variable.
- `set -x` echoes `.env` secrets into the script's output.

Reopen if you would rather rebase this onto current `main` and carry it forward instead — happy to re-review.

PR #131, opened 2025-10-28
Constraint orderby type in blocks service

glyh:
So we're less likely to accidentally cause an SQL injection bug.


dkijania (approved):

glyh:
I'm confused by the commit reverting a change on this PR that's unmerged.

PR #134, opened 2025-11-03
Remove useless filtring in events&actions sql

glyh:
This is confirmed by running this on a online archive db:
```
 select COUNT(*) from blocks where id = parent_id;
 count 
-------
     0
(1 row)
```

### DIFF
--- a/src/db/sql/events-actions/queries.ts
+++ b/src/db/sql/events-actions/queries.ts
@@ -28,7 +28,6 @@ function fullChainCTE(db_client: postgres.Sql, from?: string, to?: string) {
     FROM
       blocks b
       INNER JOIN pending_chain ON b.id = pending_chain.parent_id
-      AND pending_chain.id <> pending_chain.parent_id
       AND pending_chain.chain_status <> 'canonical'
   ), 
   full_chain AS (
@@ -42,15 +41,15 @@ function fullChainCTE(db_client: postgres.Sql, from?: string, to?: string) {
           pending_chain
         WHERE 1=1
           ${
-            // If fromAsNum is not undefined, then we have also set toAsNum and can safely query the range
-            // If no params ar provided, then we query the last BLOCK_RANGE_SIZE blocks
-            fromAsNum
-              ? db_client`AND height >= ${fromAsNum} AND height < ${toAsNum!}`
-              : db_client`AND height >= (
+    // If fromAsNum is not undefined, then we have also set toAsNum and can safely query the range
+    // If no params ar provided, then we query the last BLOCK_RANGE_SIZE blocks
+    fromAsNum
+      ? db_client`AND height >= ${fromAsNum} AND height < ${toAsNum!}`
+      : db_client`AND height >= (
                 SELECT MAX(height)
                 FROM pending_chain
             ) - ${BLOCK_RANGE_SIZE}`
-          }
+    }
         UNION ALL
         SELECT
           id, state_hash, parent_id, parent_hash, height, global_slot_since_genesis, global_slot_since_hard_fork, timestamp, chain_status, ledger_hash, last_vrf_output
@@ -59,15 +58,15 @@ function fullChainCTE(db_client: postgres.Sql, from?: string, to?: string) {
         WHERE
           chain_status = 'canonical'
           ${
-            // If fromAsNum is not undefined, then we have also set toAsNum and can safely query the range
-            // If no params ar provided, then we query the last BLOCK_RANGE_SIZE blocks
-            fromAsNum
-              ? db_client`AND b.height >= ${fromAsNum} AND b.height < ${toAsNum!}`
-              : db_client`AND b.height >= (
+    // If fromAsNum is not undefined, then we have also set toAsNum and can safely query the range
+    // If no params ar provided, then we query the last BLOCK_RANGE_SIZE blocks
+    fromAsNum
+      ? db_client`AND b.height >= ${fromAsNum} AND b.height < ${toAsNum!}`
+      : db_client`AND b.height >= (
                 SELECT MAX(b2.height)
                 FROM blocks b2
             ) - ${BLOCK_RANGE_SIZE}`
-          }
+    }
       ) AS full_chain
   )
   `;
@@ -117,10 +116,9 @@ function blocksAccessedCTE(db_client: postgres.Sql, status: BlockStatusFilter) {
     INNER JOIN full_chain b ON aa.block_id = b.id
   WHERE
     1 = 1
-    ${
-      status === BlockStatusFilter.all
-        ? db_client``
-        : db_client`AND chain_status = ${status.toLowerCase()}`
+    ${status === BlockStatusFilter.all
+      ? db_client``
+      : db_client`AND chain_status = ${status.toLowerCase()}`
     }
   )`;
 }
@@ -308,15 +306,13 @@ function emittedActionStateCTE(
       INNER JOIN zkapp_field zkf4 ON zkf4.id = zks.element4
     WHERE
       1 = 1
-    ${
-      fromActionState
-        ? db_client`AND zkf0.id >= (SELECT id FROM zkapp_field WHERE field = ${fromActionState})`
-        : db_client``
+    ${fromActionState
+      ? db_client`AND zkf0.id >= (SELECT id FROM zkapp_field WHERE field = ${fromActionState})`
+      : db_client``
     }
-    ${
-      endActionState
-        ? db_client`AND zkf0.id <= (SELECT id FROM zkapp_field WHERE field = ${endActionState})`
-        : db_client``
+    ${endActionState
+      ? db_client`AND zkf0.id <= (SELECT id FROM zkapp_field WHERE field = ${endActionState})`
+      : db_client``
     }
   )`;
 }
--- a/src/db/sql/events-actions/queries.ts
+++ b/src/db/sql/events-actions/queries.ts
@@ -28,7 +28,6 @@ function fullChainCTE(db_client: postgres.Sql, from?: string, to?: string) {
     FROM
       blocks b
       INNER JOIN pending_chain ON b.id = pending_chain.parent_id
-      AND pending_chain.id <> pending_chain.parent_id
       AND pending_chain.chain_status <> 'canonical'
   ), 
   full_chain AS (
@@ -42,15 +41,15 @@ function fullChainCTE(db_client: postgres.Sql, from?: string, to?: string) {
           pending_chain
         WHERE 1=1
           ${
-            // If fromAsNum is not undefined, then we have also set toAsNum and can safely query the range
-            // If no params ar provided, then we query the last BLOCK_RANGE_SIZE blocks
-            fromAsNum
-              ? db_client`AND height >= ${fromAsNum} AND height < ${toAsNum!}`
-              : db_client`AND height >= (
+    // If fromAsNum is not undefined, then we have also set toAsNum and can safely query the range
+    // If no params ar provided, then we query the last BLOCK_RANGE_SIZE blocks
+    fromAsNum
+      ? db_client`AND height >= ${fromAsNum} AND height < ${toAsNum!}`
+      : db_client`AND height >= (
                 SELECT MAX(height)
                 FROM pending_chain
             ) - ${BLOCK_RANGE_SIZE}`
-          }
+    }
         UNION ALL
         SELECT
           id, state_hash, parent_id, parent_hash, height, global_slot_since_genesis, global_slot_since_hard_fork, timestamp, chain_status, ledger_hash, last_vrf_output
@@ -59,15 +58,15 @@ function fullChainCTE(db_client: postgres.Sql, from?: string, to?: string) {
         WHERE
           chain_status = 'canonical'
           ${
-            // If fromAsNum is not undefined, then we have also set toAsNum and can safely query the range
-            // If no params ar provided, then we query the last BLOCK_RANGE_SIZE blocks
-            fromAsNum
-              ? db_client`AND b.height >= ${fromAsNum} AND b.height < ${toAsNum!}`
-              : db_client`AND b.height >= (
+    // If fromAsNum is not undefined, then we have also set toAsNum and can safely query the range
+    // If no params ar provided, then we query the last BLOCK_RANGE_SIZE blocks
+    fromAsNum
+      ? db_client`AND b.height >= ${fromAsNum} AND b.height < ${toAsNum!}`
+      : db_client`AND b.height >= (
                 SELECT MAX(b2.height)
                 FROM blocks b2
             ) - ${BLOCK_RANGE_SIZE}`
-          }
+    }
       ) AS full_chain
   )
   `;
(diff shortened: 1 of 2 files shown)


src/db/sql/events-actions/queries.ts
```diff
@@ -28,7 +28,6 @@ function fullChainCTE(db_client: postgres.Sql, from?: string, to?: string) {
     FROM
       blocks b
       INNER JOIN pending_chain ON b.id = pending_chain.parent_id
-      AND pending_chain.id <> pending_chain.parent_id
```
45930: I am not confident that this can never be true.  I just don't know the Mina spec well enough.  What is gained by removing this filter?  Is the performance notably better?

45930 (approved):

SanabriaRusso:
Thanks for digging into this and backing it with the `COUNT(*) … WHERE id = parent_id` = 0 check — the reasoning holds. I confirmed it against the integration fixture too: the genesis block (`id=1`, `height=1`) has `parent_id = NULL`, so the recursive walk already terminates on the NULL parent-join, and `pending_chain.id <> pending_chain.parent_id` can only ever matter for a self-parented block — of which there are none. So this is a genuine no-op on the returned events/actions rows. (No mina-explorer impact either — it doesn't use `events`/`actions`.)

Two small things before merge:

1. **Consistency:** the identical guard still lives in `getZkappsWithPendingEventsQuery` in this same file (`WHERE pc.id <> pc.parent_id AND pc.chain_status <> 'canonical'`). If it's dead code in `fullChainCTE`, it's dead in both — worth removing it there too (or keeping both) so the file doesn't contradict itself on whether the guard is needed.

2. This guard is effectively an infinite-recursion safety net — a self-parented non-canonical row would recurse forever — and it costs basically nothing to keep. Since we're now relying on the "no block has `id = parent_id`" invariant, could you drop a one-line SQL comment noting it? That way a future schema/hard-fork change that ever self-references a block surfaces in review instead of as a query hang.

Also, the branch is ~6 weeks behind `main` and shows `BLOCKED` — `queries.ts` hasn't changed on main since your last sync, so it should rebase cleanly and just needs a refresh to re-run CI.

SanabriaRusso:
**Verdict: MERGEABLE** ✅

Second pass at head `02e279d`. Both of the maintainer's asks are done: the twin guard in `getZkappsWithPendingEventsQuery` is gone (`src/db/sql/events-actions/queries.ts:460-467` at head) and both sites carry the invariant comment. The one question that matters here — *does dropping the predicate change the result set?* — resolves to **no**, and I proved it rather than took it on faith.

### What I checked

- **Join types (the LEFT-JOIN trap):** `grep -n "LEFT JOIN\|RIGHT JOIN\|FULL JOIN\|LEFT OUTER" src/db/sql/events-actions/queries.ts` returns **zero hits** — the entire file is inner joins. Both recursive terms are `INNER JOIN pending_chain ON b.id = pending_chain.parent_id` (line 36 at head) and `JOIN pending_chain pc ON b.id = pc.parent_id` (line 465 at head, plain `JOIN` = inner). So there is no nullable side, no WHERE-predicate-silently-inner-joining, and no NULL-shape change. The `ON`-vs-`WHERE` placement difference between the two sites is also semantically irrelevant for an inner join.
- **What the predicate actually constrained:** `pending_chain.id <> pending_chain.parent_id` filters the *working-table* row, not `b`. The only row it can ever exclude is one where a block is its own parent — and joining such a row produces `b.id = pc.parent_id = pc.id`, i.e. it re-emits itself, forever. It is a **cycle-breaker, not a filter**. On any DB where no block self-parents, the emitted row set is byte-identical before and after.
- **Genesis termination is unchanged and independent of the removed predicate:** `blocks.parent_id` is nullable (`tests/integration/fixtures/archive_db.sql:192`) and genesis has `parent_id = \N` (fixture blocks COPY, `id=1`, `height=1`). `b.id = NULL` evaluates to NULL → no match → recursion stops. That holds with or without the guard. (Incidentally the removed predicate *also* went NULL at genesis, so it was doubly redundant there.)
- **`chain_status` / `BlockStatusFilter` / `tokenId` are untouched:** `pending_chain.chain_status <> 'canonical'` survives at both sites, `blocksAccessedCTE(db_client, status)` and `accountIdentifierCTE(db_client, address, tokenId)` are not in the diff. CANONICAL/PENDING/ALL semantics and the token filter are bit-for-bit the same. Nothing canonical-vs-pending is at stake.
- **Downstream blast radius — neither consumer calls `events`/`actions`:** in `mina-explorer`, the only archive root fields in `src/services/api/*.ts` are `blocks(` and `networkState {`; `EventFilterOptionsInput` / `ActionFilterOptionsInput` / `events(input` / `actions(input` have **zero** hits under `src/`. In `mina-explorer-api`, the only root field across `app/upstream/*.py` + `app/indexer/*.py` is `blocks(`; even the zkApp path goes through blocks (`app/upstream/zkapp_archive.py:37` → `transactions { ... }`), never `events`/`actions`. So the only consumers of this SQL are **o1js / zkApp clients**, and since the row set is unchanged, `fetchActions` ordering and action-state reduction are unaffected.
- **`.gitignore` change is a real bug fix, not noise:** on `main`, `git check-ignore -v src/db/foo.ts` → `.gitignore:27:db/`. Any *new* file under `src/db/` is currently silently un-addable. `/db/` anchors it correctly.
- **Rebase state:** `mergeable: MERGEABLE`, `mergeStateStatus: BEHIND`. I diffed `main`'s `queries.ts` against the head blob — the only delta is this PR's own change, so the rebase is clean for this file; it just needs a refresh to re-run CI.

### Perf claim — honest answer: cosmetic

There is no plan change. The removed predicate is a per-working-row integer comparison inside a recursive term; it touches no index, changes no join order, and prunes nothing (it matches every row). @glyh already said as much in the line-31 thread ("No I don't think the perf would be much better"). The title says "useless filtring", and the correct framing is *readability + removing a contradiction between two copies of the same walk* — not speed. Worth editing the PR body to say that so nobody later cites this as a perf precedent.

### Non-blocking nits

**1. The SQL comment's justification is wrong on the FK point — and the comment is the entire warrant for this change, so it should be right.**

Head says:

> `-- (parent_id is a FK to blocks.id; a self-reference would be a cycle), so`

A self-reference is **not** a cycle a foreign key rejects. `blocks_parent_id_fkey FOREIGN KEY (parent_id) REFERENCES public.blocks(id)` (`tests/integration/fixtures/archive_db.sql:6018`) is perfectly satisfied by a row pointing at itself — Postgres validates self-referencing rows against the row's own new value, and `INSERT INTO blocks (id, parent_id, ...) VALUES (7, 7, ...)` succeeds. The guarantee is **protocol-level, not schema-level**: `parent_hash` is the parent's `state_hash`, and a block's own `state_hash` commits to `parent_hash`, so `id = parent_id` would require a hash preimage collision. The archive node also resolves `parent_id` by looking up `parent_hash`, which can never be the row being inserted.

Suggested replacement for the `fullChainCTE` comment:

```sql
      -- No id <> parent_id guard is needed. Two reasons, in order:
      --   1. The walk terminates on genesis regardless: genesis has
      --      parent_id = NULL, so `b.id = pending_chain.parent_id` is NULL and
      --      the join yields nothing.
      --   2. No block is its own parent. This is a PROTOCOL invariant, not a
      --      schema one -- blocks_parent_id_fkey does NOT reject a
      --      self-referencing row. It holds because parent_hash is the parent's
      --      state_hash and a block's state_hash commits to parent_hash, so
      --      id = parent_id would require a hash preimage collision.
      -- The dropped guard was therefore a cycle-breaker, never a filter: it
      -- could only ever exclude a self-parented row, and such a row would
      -- re-emit itself forever. If a hard fork or bulk import ever produces
      -- id = parent_id, this CTE hangs rather than returning wrong data.
      -- Pinned by the invariant test in tests/devnet-dump/devnet-dump.test.ts.
```

and shorten the `getZkappsWithPendingEventsQuery` one to `-- Same reasoning as fullChainCTE: no id <> parent_id guard needed (protocol invariant, not a FK constraint).`

**2. No existing test would catch a regression here — add the invariant test, not a query test.**

I checked what's there:
- `tests/services/events-service/events-service.test.ts` builds `ArchiveNodeDatabaseRow` objects by hand (`makeRow`, lines 7-38) and calls `blocksToEvents`. It never issues SQL. Same for `tests/services/actions-service/actions-service.test.ts`.
- The integration `EventsService` / `ActionsService` blocks (`tests/integration/integration.test.ts:276-402`) only assert **empty** results — they prove the SQL parses and runs, nothing about the pending-chain walk.

A query-level test can't help, because the change is a no-op on all well-formed data — a test asserting equal results would pass on both branches. The thing worth pinning is the **invariant this PR now depends on**. Add it to the existing `Schema validation (devnet dump)` block at `tests/devnet-dump/devnet-dump.test.ts:484`, where it runs nightly against a real chain:

```ts
  test('no block is its own parent (invariant relied on by fullChainCTE)', async () => {
    // fullChainCTE and getZkappsWithPendingEventsQuery dropped their
    // `id <> parent_id` cycle guard on the strength of this invariant. If a
    // hard fork or bulk import ever violates it, those recursive CTEs hang
    // instead of returning wrong data -- fail loudly here first.
    const rows = await client<{ id: number }[]>`
      SELECT id FROM blocks WHERE parent_id IS NOT NULL AND id = parent_id LIMIT 5
    `;
    assert.deepStrictEqual(
      rows.map((r) => r.id),
      [],
      `self-parented block(s) found: ${rows.map((r) => r.id).join(', ')} -- ` +
        'restore the `id <> parent_id` guard in src/db/sql/events-actions/queries.ts'
    );
  });

  test('exactly one root block, and it is genesis', async () => {
    const rows = await client<{ id: number; height: string }[]>`
      SELECT id, height FROM blocks WHERE parent_id IS NULL
    `;
    assert.strictEqual(rows.length, 1, 'recursion terminates on a single NULL-parent root');
    assert.strictEqual(rows[0].height, '1');
  });
```

Mirror the first one in `Schema validation (integration)` at `tests/integration/integration.test.ts:404` using `createTestClient()` so it also runs on the fixture in the fast suite. Both are pure `SELECT`s against tables already in `USED_TABLES`, so no fixture or setup change is needed.

**3. Coordination with #209 (heads-up only, no action on this PR).**

No textual conflict: #209's `queries.ts` hunks are at old lines 256 / 308 / 378 / 412-420, this PR's at 27-40 and 452-463 — they merge cleanly in either order, and #209 does not touch `fullChainCTE`'s recursion. **But** #209 adds a brand-new `WITH RECURSIVE pending_chain` inside `resolveActionStateBoundary` that carries `AND pending_chain.id <> pending_chain.parent_id`. Whichever lands second, the file goes right back to contradicting itself on whether the guard is needed — the exact thing @SanabriaRusso asked to fix in point 1. Cheapest resolution: merge this PR first, then drop that one line from #209's new CTE during its rebase. Worth a note on #209 rather than holding this one.

**4. `data/` on `.gitignore:28` has the identical bug** the PR just fixed on line 27: `git check-ignore -v tests/data/x` → `.gitignore:28:data/`. Same one-character fix (`/data/`). Fine to fold into this PR since it's already touching the adjacent line, or leave it — not a blocker either way.

SanabriaRusso (dismissed):

SanabriaRusso (dismissed):
Approving. I re-reviewed this from scratch rather than carrying forward the earlier "clean but behind" status, since `a7b01625` and the `.gitignore` change are new.

**The removed `id <> parent_id` guards are genuinely redundant.** This was the one thing that could have been a hard blocker, so I checked it rather than reasoned about it:

- In all three recursive CTEs (`src/db/sql/events-actions/queries.ts:34`, `:506`, `:585`) the guard sits on the recursive term next to `b.id = pending_chain.parent_id`, so it can only ever exclude a row that is its own parent.
- Genesis has `parent_id = NULL` — confirmed in the fixture (`tests/integration/fixtures/archive_db.sql`, block `id=1` → `\N`). For a NULL parent the join already fails on `b.id = NULL`, so the guard was never what terminated the walk.
- I scanned every `COPY public.blocks` row in the fixture for `id = parent_id`: zero matches. And a self-parent is structurally impossible — `parent_id` resolves the block whose `state_hash` equals this block's `parent_hash`, and a block's `state_hash` commits to its parent hash, so it would require a hash cycle.
- Even setting all that aside, **duplicates couldn't reach a caller**: `fullChainCTE` feeds `full_chain AS (SELECT DISTINCT …)`, `resolveActionStateBoundary` consumes `pending_chain` only via `b.id IN (…)`, and `getZkappsWithPendingEventsQuery` ends in `SELECT DISTINCT`.
- Precedent: `best_chain_til_canonical` in `src/services/blocks-service/blocks-service.ts:197` has never carried this guard. This PR makes the four chain walks consistent rather than inventing a new assumption.

**No semantic conflict with #209 (`6b72e5b`).** I diffed the whole of `queries.ts` against `main`: the only changes are the three guard removals, the comments, and formatting. #209's `fromActionState`/`endActionState` machinery — the `bounds`, `target`, and `account_action_states` CTEs and the `boundary_height` / `window_start` / `state_entering_window` projection — is byte-identical.

**The `.gitignore` fix is load-bearing, not cosmetic.** On `main`, `git check-ignore -v src/db/newfile.ts` reports `.gitignore:27:db/` — every *new* file under `src/db/` was being silently ignored, which is exactly where #182 and #187 are adding files. `.env` (`.gitignore:2`) is untouched and still ignored; nothing secret is un-ignored.

**The formatting hunks are also a fix.** The reformatted nested ternary at `queries.ts:510-513` looked like older-Prettier drift, but running the repo's pinned `prettier@3.0.3` with the repo `.prettierrc` shows the opposite: this PR's file passes `--check` and `main`'s file fails. `npm run lint` is `eslint` only, so this drift isn't caught by CI — good that it's being corrected.

**Downstream:** zero impact. No schema change (GraphQL Inspector green), no resolver-signature change, no error-text change, and neither `mina-explorer` nor `mina-explorer-api` calls the events or actions resolvers at all. `blocks-service.ts`, which does serve consumer traffic, is untouched.

Non-blocking nits:

- `queries.ts:33` says *"parent_id is a FK to blocks.id; a self-reference would be a cycle"*. A self-referencing FK doesn't actually forbid a row pointing at itself (`blocks_parent_id_fkey`, `archive_db.sql:6018`, permits it). The invariant is real, but it comes from the hash-chain construction and genesis's NULL parent, not from the constraint. Suggest rewording so the next reader doesn't rely on the FK.
- This anchors `/db/` but leaves `data/` unanchored, so the same latent trap remains for any future `**/data/` directory. #182 anchors both — worth matching.
- **Merge ordering:** #182 touches the same `.gitignore` hunk; whichever lands second needs a one-line resolution. Prefer the fully-anchored `/db/` + `/data/` form.
- The guard did give partial defence-in-depth against a 1-cycle in a malformed DB (it never covered longer cycles, e.g. A→B→A), where the CTE would now spin. Once #182 lands, `statement_timeout` bounds even that at 15 s. Not a reason to hold this.

glyh:
It's a <50 LoCs PR why it needs so many rounds of review and force pushes? I can just close it if we don't want to merge.

dkijania:
We are rebasing others so it got conflicted and then we need another review / approval.

dkijania (approved):

PR #157, opened 2026-05-05
ci: smoke load test against static archive-DB fixture

dkijania:
## Summary

Adds a smoke load test to CI. Catches pathological breakage (5xx storms, hangs, query paths that 500) without pretending to be a perf benchmark.

GitHub-hosted runners are noisy shared VMs — 10–30% run-to-run variance on throughput is normal — so any tight perf threshold would either miss real regressions in noise or flake on the runners. Loose thresholds + a short run produces a useful signal without false alarms.

**Shape:**
- Postgres 13 service container, fixture loaded from existing `tests/integration/fixtures/archive_db.sql`.
- 30-second Artillery run at 5 req/s — `benchmark/smoke.yaml`.
- Two scenarios: `events` and `actions` — the hot query paths most likely to regress.
- Loose thresholds: p95 < 5s, error-rate < 1%, every response is 200.
- Triggered on PR and on push to main. ~2 minutes wall time.
- Report + server log uploaded as artifact on both pass and fail.

**Why not `networkState`:** the static fixture has only canonical / orphaned blocks; the integration setup separately inserts a pending block to make `networkState` work. Duplicating that SQL into the workflow would be a maintenance trap. Events + actions are the regression-relevant paths anyway.

**Why not the existing `benchmark/graphql.yaml`:** it targets the wrong port (`localhost:3000` vs the real default `8080`) and references a `zkapp.csv` payload not in the repo, so it's broken for anyone who runs it today. Leaving it for a separate "fix when someone wants to actually use it" pass instead of folding two purposes into one config.

## Test plan

- [ ] Confirm the workflow runs on this PR and turns green.
- [ ] Inspect the uploaded `smoke-load-report` artifact to sanity-check the numbers (just for first-run validation, not as a regression baseline).
- [ ] Confirm wall time is roughly in the 2–3 min range; if much longer, revisit the fixture-load step.


SanabriaRusso:
Two things worth addressing before merge:

**1. Smoke-test address is probably truncated and never reaches the SQL.**

The query uses `B62qiaEMrWiYdK7LcJ2ScdMyG8LzUxi7yaw17XvBD34on7UKfhAkR` (50 chars). The fixture only contains `…UKfhAkRML` (52 chars), and real Mina B62 addresses are 55 chars. The 50-char form is almost certainly invalid base58check.

If the resolver validates the address upfront, the request fails before hitting SQL — and the smoke test's stated purpose ("the SQL doesn't blow up under load") isn't actually exercised. Suggest either using a real address from the fixture (the full `…UKfhAkRML`, or another `creator` / `pk_id`), or any syntactically valid B62. Zero rows back is fine; we just need the SQL to actually run.

While you're at it, worth doing a one-shot sanity check on the `ensure` thresholds too — set `p95: 1` deliberately, confirm CI goes red. Artillery's threshold syntax has shifted across versions and a silently-passing `ensure:` block would be the same class of bug.

**2. Workflow triggers on every PR, including docs-only.**

`on: pull_request:` with no path filter means PRs touching only `README.md` / `docs/` pay ~2 min of runner time. Cheap individually, but it adds up given the docs-only PRs currently in flight. Suggest a `paths-ignore: ['**.md', 'docs/**', 'AGENTS.md']` or an explicit `paths:` allowlist for `src/**`, `package*.json`, `schema.graphql`, the workflow, and the fixture.

Non-blocking, but easy to add now while the file is fresh.

SanabriaRusso (approved):

dkijania:
Both points addressed in a9c9f1a — plus a third (your `ensure:` sanity-check suggestion turned up something real).

**1. Truncated address.** Right call — was 50 chars, invalid base58check. Swapped for the full 55-char `B62qiaEMrWiYdK7LcJ2ScdMyG8LzUxi7yaw17XvBD34on7UKfhAkRML` from the fixture. Both event and action scenarios now use it.

**2. `ensure:` thresholds — sanity check done, found a thing.** Ran the deliberate-fail test you suggested. Couple of findings:

- The legacy v1.x shorthand (`p95: 5000`, `maxErrorRate: 1`) actually does still work in Artillery 2.0.21 — strict (`p95: 1`) exits 1, loose (`p95: 5000`) exits 0. So the original config wasn't silently passing, but it was undocumented going forward.
- Rewrote to the documented v2.x form (`thresholds:` + `conditions:`) for clarity.
- **Gotcha:** the canonical metric is `http.response_time.p95` — **not** `engine.http.response_time.p95` as in the plugin's own README example. The `engine.` prefix triggers a filtrex parse error and the check then "fails" for the wrong reason (always red regardless of actual p95). Without your suggestion to do the round-trip check, we'd have shipped this with a false-failing threshold. Documented this in a comment in the YAML.

**3. Path filter on workflow.** Added `paths:` allowlist (more conservative than `paths-ignore`) to both `pull_request` and `push: main`:
```
src/**, package.json, package-lock.json, schema.graphql,
tests/integration/fixtures/**, benchmark/smoke.yaml,
.github/workflows/smoke-load-test.yaml
```
Docs-only PRs and pushes now skip the smoke run entirely.

PR #158, opened 2026-05-06
chore: refresh stale Mina references — point to mainnet 3.3.1

dkijania:
## Summary

Several config files still pointed at pre-mainnet Berkeley-era artifacts. Most are *stale-but-running* (old image tags that no longer exist, peer list for a retired testnet) and one is *quietly wrong* (port mismatch).

| Where | Was | Now | Why |
| --- | --- | --- | --- |
| `.env.example.compose` `MINA` / `ARCHIVE` | `gcr.io/o1labs-192920/mina-{daemon,archive}:1.3.2beta2-rampup-2884275-bionic[-berkeley]` | `minaprotocol/mina-{daemon,archive}:3.3.1-7b34378-bullseye-mainnet` | Retired registry, EOL OS (bionic / Ubuntu 18.04), dead testnet (Berkeley), pre-mainnet beta. |
| `.env.example.compose` `PEER_LIST_URL` | `https://storage.googleapis.com/seed-lists/berkeley_seeds.txt` | `https://bootnodes.minaprotocol.com/networks/mainnet.txt` | Berkeley peer list is for a defunct testnet. |
| `scripts/generate_libp2p.sh` keypair image | `minaprotocol/mina-generate-keypair:1.3.0-9b0369c` | `minaprotocol/mina-daemon:3.3.1-7b34378-bullseye-mainnet` | The dedicated keypair image was retired in 2022 (last DH push 2022-05). The mina-daemon image absorbed those subcommands. |
| `scripts/generate_libp2p.sh` libp2p step | shelled out to local `mina` binary | runs in Docker | Was inconsistent — mina keypair already used Docker, libp2p didn't. Now Docker is the only requirement. |
| `.env.example.lightnet` `PORT` | `3000` | `8080` | Server default is 8080 everywhere else. Last holdout of an old port reference. |
| `tests/integration/setup.ts` header comment | claimed default dump path was `~/work/minaprotocol/mina/src/test/archive/sample_db/archive_db.sql` | matches the actual default (`tests/integration/fixtures/archive_db.sql`) | Comment was never updated when the fixture was checked in. |

**Codename choice:** `bullseye` (Debian 11) as the conservative mainnet default. mina-daemon 3.3.1 is also published for `jammy`, `focal`, `bookworm`, and `noble`. Easy to swap if a different OS family is preferred — verified via Docker Hub `tags/` API while writing this.

## Not in this PR

`tests/consensus/select-consensus-precomputed.ts` hardcodes `precomputed_berkeley/` as its input directory; that dir doesn't exist in the repo and the script is a manual run-only dev tool. Holding off on a fix here pending a decision on whether to delete the tool or repoint it at a current data source.

## Test plan

- [ ] On a clean checkout, `cp .env.example.compose .env && docker compose up` brings up the mainnet daemon (sync isn't expected to complete in CI time, but the daemon image must pull and start).
- [ ] `./scripts/generate_libp2p.sh` runs without a local `mina` binary on `$PATH`, produces both `~/keys/libp2p-keys` and the libp2p keypair files, owned by the user.
- [ ] `npm run test:integration` still passes (header-comment-only change shouldn't affect behavior, but worth confirming).
- [ ] Cross-check: the README, AGENTS.md, getting-started.md, and `.env.example.compose` now all say port 8080.


SanabriaRusso:
**Nit: `set -x` echoes the password exports**

Pre-existing, not from this PR, but since the script is already being rewritten here it's a good moment to fix.

`set -x` plus

```bash
export MINA_PRIVKEY_PASS="passlib"
export MINA_LIBP2P_PASS="passlib"
```

prints both values to stderr on every run. Harmless for `passlib` in a dev-only script, but it's a copy-paste footgun — anyone adapting this for a real credential inherits a script that logs the password by default.

Two easy options, either works:

1. Drop `set -x` — `set -eo pipefail` already covers correctness.
2. Keep it, but wrap the exports: `{ set +x; export MINA_PRIVKEY_PASS=…; export MINA_LIBP2P_PASS=…; set -x; }` and add a one-line comment that the script is dev-only and should not be adapted for real keys without disabling tracing.

If neither fits the scope, a single line in the script header (or `docs/getting-started.md`) saying "dev-only; disable `set -x` before adapting with real credentials" is enough. Non-blocking either way.

SanabriaRusso (approved):

dkijania:
Good catch — went with option 1 (drop `set -x`). The explicit `echo` lines for each step keep the run readable without tracing the password exports.

Pushed in ce43f4e.

PR #159, opened 2026-05-06
test: replace consensus port-validation dev tools with automated integration test

dkijania:
## Summary

Resolves the dead `precomputed_berkeley/` reference flagged by the stale-refs scan. Converts the manual TS-vs-OCaml port-validation harness into a real CI test that uses the existing archive-DB fixture as its source of truth.

**Removed** (`tests/consensus/`):
- `select-consensus-precomputed.ts` — referenced a `precomputed_berkeley/` block dump that doesn't exist in the repo.
- `run-compare.ts` — comparator that needed an OCaml-side output directory that doesn't exist either.
- `types.ts` — only used by the two above.

These were a one-shot harness from the original TS port; the OCaml side was produced out-of-band, never checked in, and Berkeley was retired in 2024 anyway. Unrunnable for over a year.

**Added** (`tests/integration/consensus-vs-chain.test.ts`):

Same intent (prove TS `select` matches OCaml's chain decisions) but uses the existing fixture as the reference. Every `chain_status='canonical'` block was an OCaml selection win; every `chain_status='orphaned'` was a loss. The test pulls every same-parent sibling pair from the fixture, runs `select` in both argument orderings, asserts the canonical sibling wins.

Sub-second. Runs alongside the existing integration tests.

## Why these specific filters

Two SQL filters that took a couple of investigation iterations to settle on:

**Same-parent siblings only.** Same-height blocks with *different* parents are tips of forks that converged at the same height; OCaml's decision there was driven by later length-based comparisons, not by `select` at that height. Including them would test something `select` isn't responsible for.

**Heights ≤ 2 excluded.** The static fixture is from a small bootstrap network where all nodes started simultaneously. Blocks immediately after genesis race in ways that don't reflect steady-state consensus, and produce a known mismatch at height 2 specifically. From height 3 onward the chain runs in its normal regime and `select` agrees on every sibling pair (validated locally — heights 4, 7, 10, 15, 23 all pass).

This was a real find from running the test, not a presumption. Documented in the file's docstring so future readers don't wonder.

## Coverage

5 sibling-fork groups in the static fixture, 5 canonical-vs-orphan pairs, 10 `select` calls per run (each pair in both orderings). Asserts ≥3 groups so the test fails loudly if the fixture's fork structure changes meaningfully.

## Mechanical change

`npm run test:integration` previously ran a single hardcoded file. Now globs `build/tests/integration/*.test.js`, mirroring the pattern already used by `test:unit`.

## Test plan

- [x] New test passes locally against a fresh Postgres + the static fixture.
- [x] Existing `integration.test.ts` (25 tests) still passes — 26 total when both files run via the glob.
- [x] CI confirms the same against the merge target.


SanabriaRusso:
**Worth fixing here: `test:integration` swallows mid-loop failures**

The new glob script:
```
for f in ./build/tests/integration/*.test.js; do node --enable-source-maps --test "$f"; done
```
exits with the status of the last iteration. If `consensus-vs-chain.test.js` fails (exit 1) and `integration.test.js` runs after it and passes (exit 0), the npm script exits 0 and CI stays green on a real failure.

Pre-existing in `test:unit` / `test:coverage`, but this PR is the moment it starts mattering — going from "one hardcoded file" to a glob is exactly what makes a silent-failure runner risky. Easy fix in the same script:

```bash
ec=0; for f in ./build/tests/integration/*.test.js; do node --enable-source-maps --test "$f" || ec=$?; done; exit $ec
```

Or if you want to fail fast on the first red file:

```bash
for f in ./build/tests/integration/*.test.js; do node --enable-source-maps --test "$f" || exit 1; done
```

Either is fine. Worth applying the same pattern to `test:unit` and `test:coverage` while you're in the file — same bug, same fix, and it's the kind of thing that's much easier to land as a one-line follow-up now than to explain six months from now when a flaky test masks a real one.

Non-blocking on the consensus test itself, but strongly suggest folding the script fix into this PR rather than spinning a separate one — small enough not to muddy the diff, important enough not to lose track of.

SanabriaRusso (changes requested):
https://github.com/o1-labs/Archive-Node-API/pull/159#issuecomment-4396371407

dkijania:
Good catch — folded the fix into this PR. Pushed in 2e87608:

- Aggregate `ec=$?` across iterations and `exit $ec` so a passing iteration after a failing one no longer masks the failure.
- Applied to `test:integration`, `test:unit`, and `test:coverage` while in the file.
- Wrapped in `( … )` so the leading `npm run build &&` still short-circuits cleanly (without the subshell, a build failure would skip `ec=0` and `exit $ec` would fall back to last-command status, reintroducing the same bug).

Verified all four paths in a quick bash repro: build-failure short-circuits, all-pass returns 0, mid-loop failure surfaces, and any-fail returns 1.

SanabriaRusso (approved):

PR #160, opened 2026-05-11
fix(network-service): don't crash when canonical or pending rows are missing

dkijania:
## Summary

`NetworkService.getNetworkStateData` assumed the SQL would always return one row per `chain_status`. When either side is missing — fresh archive, static test fixture, partially-loaded snapshot — `rows.filter(...)[0]` was `undefined` and accessing `.height` crashed the resolver, surfacing to clients as:

```
GraphQL error in networkState: Unexpected error.
```

Discovered while wiring up new SDKs (\`mina-archive-sdk-{js,go,rust}\`) against the existing \`tests/integration/fixtures/archive_db.sql\`: events / actions / blocks queries all worked end-to-end, only \`networkState\` blew up.

## What this PR changes

- \`src/services/network-service/network-service.ts\` — defensively handle empty / partial \`rows\`:
  - both missing → return \`{ maxBlockHeight: null }\` (the schema already allows it)
  - canonical missing → \`canonicalMaxBlockHeight = 0\` (no block finalized yet)
  - pending missing → mirror canonical (archive fully quiesced)
- \`src/blockchain/types.ts\` — widen the \`NetworkState\` type to \`maxBlockHeight: MaxBlockHeightInfo | null\` to match the schema's nullable field.
- Adjust the four existing call sites in tests that destructure \`maxBlockHeight\` to assert non-null first; they all run against fully-populated data so the assertion always holds in practice.
- New \`tests/unit/network-service.test.ts\` with 6 cases:
  - empty rows → \`maxBlockHeight=null\`
  - both present → both heights returned
  - canonical only → pending = canonical *(this is the case that previously crashed)*
  - pending only → canonical = 0
  - stringified heights coerced to numbers
  - unexpected \`chain_status\` values ignored

The \`Int!\` non-null contract on the inner fields is preserved.

## Why now

Three Archive-Node-API SDKs we're about to ship (\`o1-labs/mina-archive-sdk-{js,go,rust}\`) have integration tests that run against the same \`archive_db.sql\` fixture this repo uses. Until this lands, those SDKs have to keep a soft try/catch around \`getNetworkState\` calls in their integration suites, which masks real issues. After this merges I'll tighten those assertions back to strict.

## Test plan

- [x] \`npm run test:unit\` — 83 tests pass, including the 6 new \`NetworkService\` cases
- [x] \`npm run build\` — typecheck clean across all callers
- [x] CI: full unit + integration runs (lightnet-backed)
- [x] Manual: hit \`networkState\` against the static fixture; no longer surfaces \"Unexpected error\"

### DIFF
--- a/src/blockchain/types.ts
+++ b/src/blockchain/types.ts
@@ -25,7 +25,9 @@ export type Action = {
 };
 
 export type NetworkState = {
-  maxBlockHeight: MaxBlockHeightInfo;
+  // Nullable when the archive has no canonical or pending blocks indexed yet.
+  // Matches the schema: NetworkStateOutput.maxBlockHeight is nullable.
+  maxBlockHeight: MaxBlockHeightInfo | null;
 };
 
 export type MaxBlockHeightInfo = {
--- a/src/services/network-service/network-service.ts
+++ b/src/services/network-service/network-service.ts
@@ -32,19 +32,37 @@ class NetworkService implements INetworkService {
     sqlSpan.end();
 
     const processingSpan = tracingState.startSpan('networkState.processing');
-    const maxBlockHeightInfo = {
-      canonicalMaxBlockHeight: Number(
-        rows.filter((row) => row.chain_status === 'canonical')[0].height
-      ),
-      pendingMaxBlockHeight: Number(
-        rows.filter((row) => row.chain_status === 'pending')[0].height
-      ),
-    };
-    const networkState = {
-      maxBlockHeight: maxBlockHeightInfo
+
+    // The SQL returns at most one row per chain_status. Either row may be
+    // missing on a fresh archive or against a static fixture, so don't
+    // assume both are present.
+    const canonicalRow = rows.find((r) => r.chain_status === 'canonical');
+    const pendingRow = rows.find((r) => r.chain_status === 'pending');
+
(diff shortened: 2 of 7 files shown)


SanabriaRusso:
Nice fix — the `find()` rewrite plus the schema-aligned type widening is exactly the right shape. One small drive-by suggestion while you're in this file:

`getNetworkState` still has a leftover fallback that I think is now safely deletable:

```ts
async getNetworkState(options: unknown): Promise<NetworkState> {
  const tracingState = extractTraceStateFromOptions(options);
  return (await this.getNetworkStateData({ tracingState })) ?? [];
}
```

The `?? []` was a pre-existing oddity (returning `[]` from a function typed `Promise<NetworkState>` would have been a type lie if it ever fired), and after this PR `getNetworkStateData` always resolves to an object — either `{ maxBlockHeight: null }` or `{ maxBlockHeight: { ... } }` — so the fallback is provably unreachable.

Suggest collapsing to:

```ts
async getNetworkState(options: unknown): Promise<NetworkState> {
  const tracingState = extractTraceStateFromOptions(options);
  return this.getNetworkStateData({ tracingState });
}
```

Happy to fold it in here since the file's already open, or leave it for a follow-up if you'd rather keep this PR's diff focused on the crash fix — your call.

SanabriaRusso (approved):

PR #184, opened 2026-06-28
P0: Secure CORS default instead of '*' (#167)

### LINKED ISSUE #167: P0: Secure CORS default instead of '*'
## Problem
CORS defaults to wide-open: `origin: process.env.CORS_ORIGIN ?? '*'` in `src/server/server.ts`. Shipping `*` by default is unsafe for a public, potentially-credentialed endpoint.

## Proposal
- Default to an explicit allowlist (or deny) rather than `*`.
- Document the production `CORS_ORIGIN` value and the security rationale.

## Acceptance
- Out-of-the-box config is not `*`; operators opt into the origins they need.

---
Part of #163.

dkijania:
## What & why

Part of the production-readiness epic (#163). Closes #167.

CORS defaulted to `origin: process.env.CORS_ORIGIN ?? '*'` — out of the box the API allowed cross-origin browser access from **any** origin. For a publicly reachable endpoint, wide-open CORS should be a deliberate choice, not the fallback.

## Behavior

`CORS_ORIGIN` now resolves to a secure default:

| `CORS_ORIGIN` | Result |
| --- | --- |
| unset / empty | CORS disabled — same-origin only |
| `*` | allow any origin (explicit opt-in) |
| `https://a.com,https://b.com` | allowlist of those origins |

Server-to-server clients and `curl` are unaffected — CORS only governs browsers calling from another origin. A value that parses to no origins (e.g. just commas) falls back to the secure default rather than an empty, surprising allowlist.

## Note on behavior change

This changes the out-of-the-box default from "any origin" to "same-origin only". Deployments that rely on browser cross-origin access must now set `CORS_ORIGIN` explicitly (`*` or an allowlist). This is documented in `getting-started.md`; the Compose example keeps `*` for local-dev convenience with a comment flagging it.

## Changes

- New `src/server/cors.ts` — `resolveCorsOptions(env)` (isolated, unit-tested).
- `server.ts` uses `cors: resolveCorsOptions()`.

## Testing

- `npm run build` — clean
- `npm run test:unit` — all pass (6 new CORS cases)
- `npm run lint` — clean

### DIFF
--- a/.env.example.compose
+++ b/.env.example.compose
@@ -29,6 +29,8 @@ JAEGER=jaegertracing/all-in-one:latest
 PORT=8080
 SHUTDOWN_TIMEOUT_MS=20000
 LOG_LEVEL="info"
+# Wide-open CORS for local dev convenience. In production leave unset
+# (same-origin only) or set an explicit allowlist instead of "*".
 CORS_ORIGIN="*"
 READINESS_PING_TIMEOUT_MS=2000
 # Per-client-IP request rate limit (optional). RATE_LIMIT_MAX=0 disables it.
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -210,6 +210,7 @@ The server reads config from environment variables. `PG_CONN` is the only requir
 
 - Secure by default: when unset, the server sends no permissive CORS headers, so browsers can only call it same-origin. Server-to-server clients and `curl` are unaffected.
 - To allow browser apps on other origins, set an explicit allowlist: `CORS_ORIGIN=https://app.example.com,https://www.example.com`.
+- Allowlist entries must be exact `scheme://host[:port]` origins: no trailing slash, no wildcard subdomains, and no path.
 - `CORS_ORIGIN=*` opens the API to any origin — convenient for a fully public read API, but make it a deliberate choice rather than a default.
 
 ---
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -178,7 +178,7 @@ The server reads config from environment variables. `PG_CONN` is the only requir
 | `PORT` | `8080` | Port the GraphQL server listens on |
 | `SHUTDOWN_TIMEOUT_MS` | `20000` | Max ms to drain in-flight requests on SIGTERM before forcing exit. Keep Kubernetes `terminationGracePeriodSeconds` above this value |
 | `LOG_LEVEL` | `info` | `debug` \| `info` \| `warn` \| `error` |
-| `CORS_ORIGIN` | `*` | CORS allowed origin |
+| `CORS_ORIGIN` | *(disabled)* | Cross-origin access. Unset = same-origin only; `*` = any origin; or a comma-separated allowlist |
 | `READINESS_PING_TIMEOUT_MS` | `2000` | Upper bound on the `/readiness` database ping. Exceeding it returns 503 rather than leaving the probe to hang. Keep it below the orchestrator's probe `timeoutSeconds` |
 | `RATE_LIMIT_MAX` | `600` | Max requests per client IP per window; `0` disables rate limiting |
 | `RATE_LIMIT_WINDOW_MS` | `60000` | Rate-limit window length in milliseconds |
@@ -205,6 +205,12 @@ The server reads config from environment variables. `PG_CONN` is the only requir
 - Counting the hops: a GCP external Application Load Balancer appends two entries — the client IP, then the forwarding-rule IP — so a bare GCP LB is `TRUST_PROXY=2`, plus one for each additional in-cluster proxy. Getting this wrong fails silently in both directions: too high falls back to the socket address, too low keys on your own proxy's IP and collapses every client into one bucket. The `TRUST_PROXY=0` warning reports the observed chain length; use it to confirm.
 - The counter is in-memory and **per-instance**: with N replicas the effective limit is roughly N × `RATE_LIMIT_MAX`. A shared store (e.g. Redis) for exact cross-replica limits is tracked as deployment hardening.
 - Health checks (`/healthcheck`) are never rate-limited.
+
+### Notes on `CORS_ORIGIN`
+
+- Secure by default: when unset, the server sends no permissive CORS headers, so browsers can only call it same-origin. Server-to-server clients and `curl` are unaffected.
+- To allow browser apps on other origins, set an explicit allowlist: `CORS_ORIGIN=https://app.example.com,https://www.example.com`.
+- `CORS_ORIGIN=*` opens the API to any origin — convenient for a fully public read API, but make it a deliberate choice rather than a default.
 
 ---
 
--- a/src/server/cors.ts
+++ b/src/server/cors.ts
@@ -1,0 +1,42 @@
+export { resolveCorsOptions };
+export type { CorsResult };
+
+const CORS_METHODS = ['GET', 'POST'];
+
+/**
+ * Either a Yoga CORS configuration object or `false` to disable CORS entirely
+ * (no `Access-Control-Allow-Origin` header → same-origin only).
+ */
+type CorsResult = false | { origin: string | string[]; methods: string[] };
+
+/**
+ * Resolve the Yoga `cors` option from `CORS_ORIGIN`.
+ *
+ * Secure by default: when `CORS_ORIGIN` is unset, cross-origin browser access is
+ * disabled rather than allowed from anywhere. Wide-open access (`*`) is still
+ * available, but only as a deliberate opt-in. Server-to-server clients and curl
+ * are unaffected by CORS; this only governs browsers calling from another origin.
+ *
+ * Accepted values:
+ *   - unset / empty  → `false` (same-origin only)
+ *   - `*`            → allow any origin (explicit opt-in)
+ *   - one or more comma-separated origins → allowlist
+ */
+function resolveCorsOptions(
+  env: Record<string, string | undefined> = process.env
+): CorsResult {
+  const raw = env.CORS_ORIGIN?.trim();
+  if (!raw) return false;
+  if (raw === '*') return { origin: '*', methods: CORS_METHODS };
+
+  const origins = raw
+    .split(',')
+    .map((origin) => origin.trim())
+    .filter((origin) => origin.length > 0);
+
+  // A value that was non-empty but parsed to nothing (e.g. just commas) falls
+  // back to the secure default rather than an empty, surprising allowlist.
+  if (origins.length === 0) return false;
+
+  return { origin: origins, methods: CORS_METHODS };
+}
--- a/src/server/cors.ts
+++ b/src/server/cors.ts
@@ -2,12 +2,31 @@ export { resolveCorsOptions, warnIfCorsDisabled };
 export type { CorsResult };
 
 const CORS_METHODS = ['GET', 'POST'];
+// Pinned rather than echoed from Access-Control-Request-Headers: echoing makes
+// the adapter drop Vary: Origin on preflight responses while ACAO is per-origin.
+const CORS_HEADERS = ['content-type'];
 
 /**
  * Either a Yoga CORS configuration object or `false` to disable CORS entirely
  * (no `Access-Control-Allow-Origin` header → same-origin only).
  */
-type CorsResult = false | { origin: string | string[]; methods: string[] };
+type CorsResult =
+  | false
+  | {
+      origin: string | string[];
+      methods: string[];
+      allowedHeaders: string[];
+      // This API takes no cookies or Authorization header. Leaving this
+      // undefined makes the adapter send Access-Control-Allow-Credentials: true
+      // alongside an echoed origin.
+      credentials: false;
+    };
+
+const corsBase = {
+  methods: CORS_METHODS,
+  allowedHeaders: CORS_HEADERS,
+  credentials: false,
+} as const;
 
 /**
  * Resolve the Yoga `cors` option from `CORS_ORIGIN`.
@@ -27,7 +46,7 @@ function resolveCorsOptions(
 ): CorsResult {
   const raw = env.CORS_ORIGIN?.trim();
   if (!raw) return false;
-  if (raw === '*') return { origin: '*', methods: CORS_METHODS };
+  if (raw === '*') return { origin: '*', ...corsBase };
 
   const origins = raw
     .split(',')
@@ -38,7 +57,22 @@ function resolveCorsOptions(
   // back to the secure default rather than an empty, surprising allowlist.
   if (origins.length === 0) return false;
 
-  return { origin: origins, methods: CORS_METHODS };
+  for (const origin of origins) {
+    if (
+      origin.endsWith('/') ||
+      origin !== origin.toLowerCase() ||
+      origin.includes('*')
+    ) {
+      console.warn(
+        `[cors] CORS_ORIGIN entry "${origin}" is unlikely to ever match: ` +
+          "origins are compared byte-for-byte against the browser's Origin " +
+          'header, which is lowercase scheme://host[:port] with no trailing ' +
+          'slash. Wildcard subdomains are not supported.'
+      );
+    }
+  }
+
+  return { origin: origins, ...corsBase };
 }
 
 /**
@@ -47,7 +81,7 @@ function resolveCorsOptions(
  * When CORS is disabled, a browser client's requests fail in the browser — the
  * server answers normally and logs nothing, so the only symptom is a blank page
  * on someone else's screen. Saying so once at boot turns that into a glance at
- * the logs. Kept separate from `resolveCorsOptions` so that stays pure.
+ * the logs.
  */
 function warnIfCorsDisabled(
   cors: CorsResult,
--- a/src/server/cors.ts
+++ b/src/server/cors.ts
@@ -1,4 +1,4 @@
-export { resolveCorsOptions };
+export { resolveCorsOptions, warnIfCorsDisabled };
 export type { CorsResult };
 
 const CORS_METHODS = ['GET', 'POST'];
@@ -39,4 +39,25 @@ function resolveCorsOptions(
   if (origins.length === 0) return false;
 
   return { origin: origins, methods: CORS_METHODS };
+}
+
+/**
+ * Announce the secure default at startup.
+ *
+ * When CORS is disabled, a browser client's requests fail in the browser — the
+ * server answers normally and logs nothing, so the only symptom is a blank page
+ * on someone else's screen. Saying so once at boot turns that into a glance at
+ * the logs. Kept separate from `resolveCorsOptions` so that stays pure.
+ */
+function warnIfCorsDisabled(
+  cors: CorsResult,
+  warn: (message: string) => void = console.warn
+): void {
+  if (cors !== false) return;
+  warn(
+    '[cors] CORS_ORIGIN is unset — cross-origin browser requests are blocked ' +
+      '(same-origin only), and they will fail silently from this server\'s side. ' +
+      'Set CORS_ORIGIN to your frontend origin(s), or to "*" for a public API ' +
+      'any browser may call.'
+  );
 }
--- a/src/server/server.ts
+++ b/src/server/server.ts
@@ -4,6 +4,7 @@ import { Plugin } from '@envelop/core';
 import { schema } from '../resolvers.js';
 import type { GraphQLContext } from '../context.js';
 import { useReadiness } from './readiness.js';
+import { resolveCorsOptions } from './cors.js';
 
 export {
   BLOCK_RANGE_SIZE,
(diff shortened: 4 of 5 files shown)


SanabriaRusso:
Nice — secure-by-default CORS is the right call, and I like that `resolveCorsOptions` is a pure, env-injectable function with its own unit tests (the "just commas → `false`" edge case is a good touch).

One friendly backwards-compat heads-up: the [mina-explorer](https://github.com/o1-labs/mina-explorer) is a browser app that calls the archive endpoint **cross-origin**, so it relies on CORS. After this change, an archive that doesn't set `CORS_ORIGIN` will block the Explorer (and any other browser client) — and that failure is only visible in the browser; the server logs stay silent. Deployments that already set `CORS_ORIGIN` explicitly are unaffected, so this isn't breaking in practice — but a self-hosted operator upgrading on the old permissive default could hit a confusing "blank Explorer" with nothing in the logs.

Since "secure by default" is the right goal, rather than loosen it, it might help to make the new default *diagnosable* with a one-line startup warning:

```ts
const cors = resolveCorsOptions();
if (cors === false) {
  console.warn(
    '[cors] CORS_ORIGIN is unset — cross-origin browser requests are blocked (same-origin only). ' +
    'Set CORS_ORIGIN to your frontend origin(s), or "*", to allow them.',
  );
}
// ...pass `cors` into createYoga(...)
```

Not a blocker — it just turns a "why is the Explorer showing blank pages?" debugging session into a glance at the logs. 🙂

dkijania:
Thanks @SanabriaRusso — added in `6785598`, essentially your snippet.

The framing that sold it: the failure isn't just confusing, it's *invisible from the server*. The request is answered normally, nothing is logged, and the only symptom is a blank page on someone else's screen — so the one place an operator would look has nothing to say. It now announces itself once at boot.

Two small deviations:

1. **Kept `resolveCorsOptions` pure.** The warning is a separate `warnIfCorsDisabled(cors, warn)` that `buildServer` calls, with an injected sink — so it's unit-tested rather than just written, and the pure env-injectable resolver you liked stays that way.

2. **It also fires when `CORS_ORIGIN` was set but parsed to no origins** (the `', ,'` case). That's the worst version of this: the operator believes they configured an allowlist, and it silently denies every browser. Same symptom, no clue — worth the same warning.

Docs and the deployment guidance in #186 now explain when `*` is the right answer, so the warning points somewhere useful rather than just nagging.

SanabriaRusso:
**Verdict: MERGEABLE** ✅

Second pass focused on the browser consumer (`mina-explorer`, a Vite SPA that POSTs cross-origin to `https://archive-node-api.gcp.o1test.net` and friends with no backend proxy).

### What I checked

- **The maintainer's ask is actually addressed.** `warnIfCorsDisabled` exists at head (`src/server/cors.ts:52-63`) and `buildServer` calls it before `createYoga` (`src/server/server.ts:16-17`). It also fires on the `', ,'` → `false` path, which is the nastier case. Nothing to re-raise.
- **The default change does not take the Explorer down.** The Explorer's only requirement is that the deployed endpoints keep returning an ACAO header; it sends `Content-Type: application/json` with no credentials (`mina-explorer` `src/services/api/client.ts:40-42`) against the four endpoints in `src/config/networks.ts:38,47,55,69`. Every o1-labs deployment of this image sets `CORS_ORIGIN: "*"` explicitly — it's the chart default in the `archive-node-api` chart values *and* repeated in each per-network override (mainnet, devnet, pre-mesa, mesa-rc, mesa). So the new "unset ⇒ disabled" default is never reached in our deployments. Together with the documented env var (`docs/getting-started.md:180,195-199`), the boot warning, and the Compose comment (`.env.example.compose:32-34`), this clears HARD CONSTRAINT #2's exception criteria.
- **Preflight still works on the configured path.** I ran the real `graphql-yoga@4.0.4` / `@whatwg-node/server@0.9.25` in-process against each `resolveCorsOptions` output. With `CORS_ORIGIN=*`: `OPTIONS` → `204`, `access-control-allow-methods: GET, POST`, `access-control-allow-headers: content-type`, `access-control-allow-origin: <echoed origin>`. With a 2-origin allowlist and a matching `Origin`: same, plus `vary: Origin` on the POST. No 405, no rate-limiter interaction (the CORS plugin's `onRequest` short-circuits `OPTIONS` before anything else).
- **`mina-explorer-api` is unaffected** — httpx is server-side, CORS never applies.
- **Origin matching is exact-string** against the browser's `Origin` value (`@whatwg-node/server/cjs/plugins/useCors.js`, `corsOptions.origin.includes(currentOrigin)`), which is the correct semantics; the comma-split + trim in `resolveCorsOptions` produces exactly that. No off-by-one in the parsing.

### Non-blocking nits

**1. Two library footguns that a two-line addition closes — worth doing *here* since this is the CORS PR.**

Both are pre-existing on `main` (same `*` value), so neither blocks, but they're squarely in scope:

- `@whatwg-node/server` sets `Access-Control-Allow-Credentials: true` whenever `credentials` is left `undefined` and the emitted ACAO isn't the literal `*` — and with `origin: '*'` it *echoes* the caller's origin rather than emitting `*`, so today prod answers every origin with `ACAO: <you>` **plus** `ACAC: true`. Harmless while there is no auth/cookie, but it's credentialed CORS for the entire internet, which is odd in a PR titled "secure CORS".
- Worse: on the **preflight**, the library builds `Vary: Origin` and then overwrites it with `Vary: Access-Control-Request-Headers` (literal `headers['Vary'] = ...` after the `+=` branch, same file). So the 204 preflight is emitted with **no `Vary: Origin`** while `ACAO` is per-request — a shared CDN/ingress in front of the API can serve one origin's preflight to everyone. Confirmed in the probe: `vary: Access-Control-Request-Headers` only.

Pinning `allowedHeaders` and `credentials` fixes both (verified: preflight then comes back with `vary: Origin` and no `ACAC`):

```diff
--- a/src/server/cors.ts
+++ b/src/server/cors.ts
@@
 const CORS_METHODS = ['GET', 'POST'];
+// Pinned rather than echoed from `Access-Control-Request-Headers`: echoing makes
+// the adapter overwrite `Vary: Origin` with `Vary: Access-Control-Request-Headers`
+// on the preflight, so a CDN can cache one origin's 204 for every origin.
+const CORS_HEADERS = ['content-type'];
 
-type CorsResult = false | { origin: string | string[]; methods: string[] };
+type CorsResult =
+  | false
+  | {
+      origin: string | string[];
+      methods: string[];
+      allowedHeaders: string[];
+      // This API takes no cookies or Authorization header. Leaving this
+      // undefined makes the adapter send `Access-Control-Allow-Credentials: true`
+      // alongside an echoed origin.
+      credentials: false;
+    };
+
+const corsBase = { methods: CORS_METHODS, allowedHeaders: CORS_HEADERS, credentials: false } as const;
@@
-  if (raw === '*') return { origin: '*', methods: CORS_METHODS };
+  if (raw === '*') return { origin: '*', ...corsBase };
@@
-  return { origin: origins, methods: CORS_METHODS };
+  return { origin: origins, ...corsBase };
```

Test that would catch a regression (add to `tests/unit/cors.test.ts`) — it drives a real Yoga instance and asserts the *emitted preflight headers*, not the config object, so it also fails if the adapter's behaviour changes under us. Only imports `graphql-yoga`, which is already a direct dependency:

```ts
import { createYoga, createSchema } from 'graphql-yoga';

test('preflight: 204, POST + content-type allowed, Vary: Origin, no credentials', async () => {
  const yoga = createYoga({
    schema: createSchema({ typeDefs: 'type Query { hello: String }' }),
    logging: false,
    cors: resolveCorsOptions({
      CORS_ORIGIN: 'https://a.example.com,https://b.example.com',
    }),
  });
  const res = await yoga.fetch('http://localhost/', {
    method: 'OPTIONS',
    headers: {
      origin: 'https://a.example.com',
    },
  });
  assert.strictEqual(res.status, 204);
  assert.strictEqual(
    res.headers.get('access-control-allow-origin'),
    'https://a.example.com'
  );
  assert.match(res.headers.get('vary') ?? '', /Origin/);
  assert.match(res.headers.get('access-control-allow-methods') ?? '', /POST/);
  assert.match(res.headers.get('access-control-allow-headers') ?? '', /content-type/);
  assert.strictEqual(res.headers.get('access-control-allow-credentials'), null);
});
```

Without the diff above, that test fails on `vary` (`Access-Control-Request-Headers` only) and on `access-control-allow-credentials` (`"true"`) — I ran both variants against `graphql-yoga@4.0.4` to confirm it is red before and green after.

(Update the existing `deepStrictEqual` cases to include the two new keys.)

**2. `OPTIONS` returns `405` when CORS is disabled.** Yoga skips the CORS plugin entirely for `cors: false` (`graphql-yoga/cjs/server.js:171`: `options?.cors !== false && useCORS(options?.cors)`), so `OPTIONS /` falls through to `405 Allow: GET, POST` instead of a 204. Consistent with "no cross-origin", and browsers fail the preflight either way — just be aware if any probe/ingress ever does an `OPTIONS` liveness check.

**3. Allowlist entries are matched byte-for-byte, and near-misses are silent.** `https://app.example.com/` (trailing slash), `HTTPS://App.Example.com`, or a wildcard-subdomain attempt like `*.example.com` all parse "successfully", produce no warning, and then never match — the operator gets the exact blank-page-with-silent-logs failure this PR's warning was written to prevent. Cheap guard inside `resolveCorsOptions`, after the filter:

```ts
for (const origin of origins) {
  if (origin === '*') continue;
  if (origin.endsWith('/') || origin !== origin.toLowerCase() || origin.includes('*')) {
    console.warn(
      `[cors] CORS_ORIGIN entry "${origin}" is unlikely to ever match: origins are ` +
        'compared byte-for-byte against the browser\'s Origin header, which is ' +
        'lowercase scheme://host[:port] with no trailing slash. Wildcard subdomains ' +
        'are not supported.'
    );
  }
}
```
Worth one line in `docs/getting-started.md` too: entries must be exact `scheme://host[:port]`, no trailing slash, no wildcard subdomains.

SanabriaRusso (dismissed):

SanabriaRusso (dismissed):
**Approving.** This is a genuine security fix and it is implemented correctly against the adapter that actually ships. Flagging one deployment prerequisite that must not get lost in the merge train.

**Deployment prerequisite — this WILL break mina-explorer if rolled out without a config change**

mina-explorer's browser SPA calls this API directly and cross-origin (`.env.development`: `VITE_API_ENDPOINT=https://archive-node-api.mesa-rc.minaprotocol.com`, POSTed from `src/services/api/client.ts:39`). With `CORS_ORIGIN` unset, `src/server/cors.ts:48` returns `false` and `graphql-yoga/cjs/server.js:171` (`options?.cors !== false && useCORS(...)`) skips the CORS plugin entirely — no `Access-Control-Allow-Origin` on any response — so every explorer query fails as a browser `TypeError`. The archive client has no `isCorsError` branch (that exists only for the daemon endpoint, `src/services/api/daemon.ts:88`), so the UI surfaces a generic network error with no clue as to the cause.

mina-explorer-api is unaffected — it is a server-side httpx client and CORS does not apply.

That is the intended behaviour of the PR, not a defect, and the mitigations are about as good as a config-required breaking change gets: the docs table now reads *(disabled)*, there is a "Notes on `CORS_ORIGIN`" section, `warnIfCorsDisabled` (`cors.ts:86`) logs at boot, and #186 dedicates a "CORS and browser clients" section that names mina-explorer explicitly and states `CORS_ORIGIN=*` is the correct setting for a public read API. **Requirements: set `CORS_ORIGIN` in the same rollout, and list this in #198's breaking-change categories and in #197's runbook.**

**This also fixes a live hole on `main`, which strengthens the case for it**

Running the current `main` config (`origin: '*'`, `credentials` undefined) through `@whatwg-node/server/cjs/plugins/useCors.js`: the adapter reflects the caller's origin into ACAO and then takes `else if (headers['Access-Control-Allow-Origin'] !== '*') { headers['Access-Control-Allow-Credentials'] = 'true' }`. Today's server answers `acao: https://evil.example, acac: true` for **any** origin. The explicit `credentials: false` at `cors.ts:22,28` suppresses that. Real fix, correctly reasoned in the code comment.

**Verified**

- `allowedHeaders: ['content-type']` (`cors.ts:7`) covers both consumers: mina-explorer sends only `Content-Type: application/json` (`client.ts:41`, `accounts.ts:117`, `daemon.ts:37,66`) with no credentials.
- The pinning rationale at `cors.ts:5-6` is accurate against the shipped yoga 4.0.4: `useCors.js` does `headers['Vary'] = 'Access-Control-Request-Headers'` — an unconditional **overwrite** that discards `Origin` — whenever `allowedHeaders` is empty and the request echoes headers. Pinning takes the other branch and preserves `Vary: Origin`. Confirmed `Vary: Origin` + `ACAO: <origin>` + no `Allow-Credentials` on preflight for a multi-origin allowlist.
- No substring/prefix matching bug: the adapter matches with `Array.prototype.includes(currentOrigin)` — exact. Empirically, allowlist `['https://example.com','https://b.example.com']` + `Origin: https://evil-example.com` → `ACAO: "null"`, blocked.
- Plugin ordering unchanged (`server.js:145-215`: health check `:166`, CORS `:171`, user plugins `:215`).

**Non-blocking**

1. **A bare `*` inside a multi-value allowlist silently opens CORS to every origin, and the warning says the opposite.** `CORS_ORIGIN="https://app.example.com,*"` skips the `raw === '*'` branch at `cors.ts:49`, yields `origin: ['https://app.example.com','*']`, and the adapter's `corsOptions.origin.includes('*')` then reflects **any** origin — observed `acao: https://evil.example, vary: Origin`. Meanwhile `cors.ts:64-66` warns that the entry "is unlikely to ever match", which is actively misleading in a security-relevant way. Suggest treating it as the wide-open case:

```ts
const origins = raw.split(',').map((o) => o.trim()).filter((o) => o.length > 0);
if (origins.length === 0) return false;
// A bare "*" anywhere in the list opens CORS to every origin in the adapter,
// so surface it as the deliberate wide-open case rather than a dud entry.
if (origins.includes('*')) return { origin: '*', ...corsBase };
```

2. **Add a negative allowlist test** — exact matching is the security property and nothing in `tests/unit/cors.test.ts` pins it:

```ts
test('an allowlist entry is not a prefix or substring match', async () => {
  const yoga = createYoga({
    schema: createSchema({ typeDefs: 'type Query { hello: String }' }),
    logging: false,
    cors: resolveCorsOptions({ CORS_ORIGIN: 'https://example.com,https://b.example.com' }),
  });
  const res = await yoga.fetch('http://localhost/', {
    method: 'OPTIONS',
    headers: {
      origin: 'https://evil-example.com',
    },
  });
  assert.strictEqual(res.headers.get('access-control-allow-origin'), 'null');
});
```

3. **`Vary` behaviour differs by allowlist size.** A single-entry allowlist takes the adapter's `origin.length === 1` branch: constant ACAO, no `Vary: Origin` — harmless, the response genuinely does not vary. But the *non-matching* multi-entry case emits `ACAO: "null"` also with no `Vary`, so a shared cache keyed without `Origin` could serve that rejection to a legitimately allowlisted origin. Adapter behaviour rather than yours — just make sure the CDN/LB varies on `Origin`.
4. **Interaction with #185, worth not decoupling later.** With `cors: false` the CORS plugin is not registered, so its OPTIONS `stopEarly()` does not exist and #185's own `if (request.method === 'OPTIONS') return;` becomes load-bearing rather than belt-and-braces. It is present in #185 today. This matters more than it looks: mina-explorer sends `Content-Type: application/json`, which is not CORS-safelisted, so every one of its GraphQL POSTs is preceded by a preflight.
5. **Interaction with #190:** pinning `allowedHeaders` means that if anyone later wants browser clients to send `x-request-id`, preflight will reject it until `content-type, x-request-id` is added here. Neither consumer sends it today, so no action needed now.

SanabriaRusso (dismissed):

Re-verified: the PR content at `1a2d354` is **byte-identical** to the commit I approved (`d6abfce`) once both are diffed against their respective merge bases. The rebase onto `7361526` was mechanical.

- This is a **deployment prerequisite, not a defect.** It breaks mina-explorer unless `CORS_ORIGIN` is set in the same rollout — confirmed by reading the SPA, not inferred: `.env.development` sets `VITE_API_ENDPOINT` to the archive-node host and it is called cross-origin from the browser. With `cors:false` yoga does not register `useCORS` at all, so there is no ACAO on any response. #196's manifests set `CORS_ORIGIN='*'`, so the reference deployment is correct.
- It also **fixes a live hole on main**: today's config reflects any origin into ACAO *and* sets `Access-Control-Allow-Credentials: true`.
- Mitigations are good — docs, startup warning, and #186 names the consumer.

**Non-blocking:**

- A bare `'*'` inside a multi-value allowlist silently opens CORS to every origin, while the code warns it "is unlikely to ever match".
- No negative allowlist test.
- mina-explorer's `src/services/api/daemon.ts:36` sends `Content-Type: application/json`, which is not CORS-safelisted, so **every** GraphQL POST is preceded by an OPTIONS preflight. Preflight is a high-volume real path here, not an edge case.

SanabriaRusso (approved):

### Delta since the approved commit: rebase + one adaptation to merged siblings

Approved SHA `1a2d354` (merge-base `7361526`) → current head `7eb3c2d` (merge-base `3fd3379`). Diffing each against its own merge-base, the only non-context change is in `src/server/server.ts`: main split `buildServer` into `buildYoga` + `buildServer` (for #195's error-masking tests), so this PR now sets `cors` inside `buildYoga` and calls `warnIfCorsDisabled(resolveCorsOptions())` from `buildServer`. Same behaviour, correctly placed — the warning still fires once per process at boot, and `buildYoga` (which the tests drive directly) still gets the resolved `cors`. `src/server/cors.ts` and `tests/unit/cors.test.ts` are unchanged. Category: **(b) rebase + adaptation to merged siblings**.

`package-lock.json` at this head is **byte-identical** to `origin/main`'s (1579 entries / 1579 `"resolved"` / 1579 `"integrity"`), so the #190-style lockfile regression cannot apply.

### Re-verified by execution at `7eb3c2d`

`npm ci && npm run build && npm run test:unit` → build clean, **143/143 unit tests pass**, `npm run lint` exit 0. CI: all 9 checks pass.

**Preflight under yoga 5, through the real `buildYoga`, with mina-explorer's exact request shape** (`OPTIONS` + `Origin` + `Access-Control-Request-Method: POST` + `Access-Control-Request-Headers: content-type`, because `daemon.ts:36` sends `Content-Type: application/json` so *every* GraphQL POST is preflighted):

```
CORS_ORIGIN="*"    OPTIONS 204  ACAO=<caller origin>  ACAM="GET, POST"  ACAH="content-type"  ACAC=null  Vary="Origin"
                   POST    200  ACAO=<caller origin>  ACAC=null
CORS_ORIGIN=<origin>  OPTIONS 204  ACAO=<that origin>  ACAM="GET, POST"  ACAH="content-type"  ACAC=null
CORS_ORIGIN unset     OPTIONS 405  ACAO=null                      <- the intended secure default
```

And the hole this closes, replicated verbatim from `origin/main:src/server/server.ts:36-39` with `CORS_ORIGIN` unset:

```
MAIN TODAY: OPTIONS 204  ACAO=<any caller origin>  Access-Control-Allow-Credentials: true
```

Origin reflection *plus* `credentials: true` on today's default is a real hole; this PR removes it.

### The deployment prerequisite still holds after the merged siblings

Re-checked #196's manifests on `origin/main`: `deploy/kubernetes.yaml:65` sets `CORS_ORIGIN: '*'` and `deploy/docker-compose.prod.yml:23` sets `CORS_ORIGIN: '*'`, both with a comment naming the exact failure mode ("Browser clients … CANNOT reach this API unless their origin is allowed here — and they fail silently"). So the reference deployment stays correct the moment this merges. `docs/versioning.md:144` also says browser deployments must set it deliberately. **The requirement is unchanged: whatever rollout ships this must set `CORS_ORIGIN`, or mina-explorer goes blank with nothing in the server logs.**

### Interaction with the sibling PRs that landed since round 3

Short-circuited responses still carry CORS headers under yoga 5, so browser clients see the real status instead of an opaque CORS failure:

```
rate-limit (#185) 429  ACAO=<caller origin>
readiness  (#187) 503  ACAO=<caller origin>
```

### Non-blocking

- Confirmed round 3's finding rather than assuming it: `CORS_ORIGIN='https://a.example,*'` **does** reflect an unrelated origin (`OPTIONS 204, ACAO=https://totally-unrelated.example`) while the code logs that `*` "is unlikely to ever match". Misleading message, not an escalation — `*` is already a supported explicit value — but worth a follow-up.
- A non-matching origin against a single-entry allowlist returns `204` with `ACAO` set to the *allowed* origin rather than omitting the header. The browser still blocks it (exact-match rule), and with a constant ACAO the missing `Vary: Origin` is correct, so this is cosmetic.
- `src/server/cors.ts` and `tests/unit/cors.test.ts` are the only two files this PR adds that fail the repo's pinned `prettier@3.0.3 --check`. Not gating — CI's Linting job runs `--debug-check`, and `main` already fails `--check` on 42 files — but a `npx prettier --write` on those two would keep the new files clean.

Issue #173, opened 2026-06-28
P1: DB pool-saturation metrics (RED metrics landed in #191)


dkijania:
## Problem
There is no metrics endpoint — only optional Jaeger tracing. We can't see request rate, error rate, latency distribution, or DB pool saturation in production.

## Proposal
- Expose Prometheus `/metrics` with RED metrics (request rate, errors, duration histogram) and DB pool saturation gauges.
- Document scrape config.

## Acceptance
- `/metrics` exposes RED + DB pool metrics suitable for alerting/dashboards.

---
Part of #163.

dkijania:
## Re-scoped after #191

#191 merged as `Refs`, not `Closes`, so this issue stays open for the part it did not deliver. Retitled to match what is actually left.

**Landed in #191** — `src/server/metrics.ts` exposes the RED set at `/metrics` behind `ENABLE_METRICS`:

- `http_requests_total` (method, path, status)
- `http_request_duration_seconds` histogram
- `http_requests_in_flight` gauge

**Still open — DB pool saturation.** `postgres.js` exposes no pool-introspection API, so there is no pool object to read `total` / `idle` / `waiting` from. Getting these needs a query-instrumentation pass rather than a gauge that samples pool state: wrap the client so acquire/release and queue depth are observable, then export from there.

Suggested acceptance criteria:

- [ ] A gauge for connections in use vs. `PG_MAX_CONNECTIONS`, so saturation is visible before it becomes latency.
- [ ] A counter or histogram for time spent waiting for a connection.
- [ ] Unit coverage that the gauges move under a saturated pool.

Related: #191, #182 (which set `PG_MAX_CONNECTIONS`, the denominator these gauges need).

PR #191, opened 2026-06-28
P1: Expose Prometheus metrics at /metrics (#173)

dkijania:
## What & why

Part of the production-readiness epic (#163). Refs #173.

There was no metrics endpoint — only optional Jaeger tracing — so request rate, errors, latency, and process health were invisible in production.

## Changes

- New `/metrics` endpoint (prom-client) via a Yoga plugin.
- **RED HTTP metrics**: `http_requests_total{method,route,status}`, `http_request_duration_seconds` histogram, `http_requests_in_flight` gauge.
- **Node process metrics**: CPU, memory, event loop lag, GC.
- Route labels normalised to a known set (`/`, `/healthcheck`, `/readiness`, `/metrics`; anything else → `other`) to **bound label cardinality**; the `/metrics` scrape is not self-counted.

New dependency: `prom-client`.

## On DB pool saturation

The issue also lists DB pool-saturation gauges. `postgres.js` exposes **no pool-introspection API**, so those need a query-instrumentation pass rather than a simple read of pool state. I scoped that out of this PR and left **#173 open** for it (this PR is `Refs`, not `Closes`) — happy to follow up with an instrumentation approach if you want it.

## Testing

- `npm run build` — clean
- `npm run test:unit` — all pass; tests cover the exposition format, per-route/status counting (asserts a request count of 2), and exclusion of the scrape endpoint


SanabriaRusso:
Thanks for this — a `/metrics` endpoint with `route` normalized to a known set and the scrape self-excluded is exactly the right shape, and bounding label cardinality up front is great to see. Two small, non-blocking operational notes:

**1. `/metrics` is always-on and unauthenticated on the public port.** `plugins.push(useMetrics())` is unconditional, so `/metrics` is reachable by anyone who can hit the GraphQL endpoint on `:8080` — which is public (CORS defaults to `*` for the mina-explorer browser client). The RED metrics are safe aggregates, but `collectDefaultMetrics` also publishes `nodejs_version_info`, process start time, and memory/GC shape. Every other optional surface here is env-gated (`ENABLE_GRAPHIQL`, `ENABLE_INTROSPECTION`, `ENABLE_LOGGING`), so it'd be consistent to gate this too and/or document the boundary (#173 also asks to "document scrape config"):

```ts
if (process.env.ENABLE_METRICS === 'true') plugins.push(useMetrics());
```
> Docs note: `/metrics` shares the API port and is unauthenticated — restrict it to your Prometheus scrapers at the ingress/LB.

**2. Heads-up for #185 (rate limiting):** its limiter currently exempts only `/healthcheck`. Since Prometheus scrapes `/metrics` on a fixed interval, it's worth adding `/metrics` to that exemption list. Note both PRs insert their plugin at the top of `buildPlugins()`, so expect a conflict there — resolve it so the plugin that short-circuits first is intentional.

And +1 on leaving #173 open for the DB-pool gauges — `postgres.js` has no pool introspection, so that genuinely needs a separate instrumentation pass.

SanabriaRusso:
**Verdict: MERGEABLE** ✅

Second pass at `bacb182`. @SanabriaRusso's point 1 is addressed: `useMetrics()` is now gated behind `ENABLE_METRICS === 'true'` (`src/server/plugins.ts:16-19`), documented in `README.md`, `docs/getting-started.md` (incl. the "shares the API port and is unauthenticated — restrict it at the ingress or load balancer" note), `.env.example.compose`, `src/envionment.d.ts`, and covered by a test (`tests/unit/metrics.test.ts:42-65`). Point 2 was about #185, so it still stands over there.

### What I checked

- **HARD CONSTRAINT #1 (error text) — safe.** The plugin only implements `onRequest`/`onResponse`; no `onParse`/`onValidate`/`onExecute`, no `maskedErrors` change. I ran a Yoga 4.0.4 server with an equivalent plugin loaded and confirmed the body is byte-identical: `{"errors":[{"message":"Cannot query field \"protocolState\" on type \"Query\".",...}]}`. So `mina-explorer-api`'s `SCHEMA_ERROR_MARKERS` tier fallback (`app/upstream/graphql.py`) and `mina-explorer`'s `error.message.includes('inBestChain')` (`src/services/api/bestChainFilter.ts`) are untouched.
- **No unbounded / client-controlled label.** Labels are `method`, `route`, `status`. `route` is allowlisted to 5 values via `KNOWN_ROUTES` (`src/server/metrics.ts:16,56-64`) and the query string is dropped. Nothing carries operation name, query text, variables, IP, or UA — so no cardinality DoS and no request-content leak. `method` is attacker-chosen but bounded: Node's HTTP parser 400s anything outside `http.METHODS` (verified: 35 entries, `FOOBARBAZ` → `400 Bad Request` before the listener runs).
- **Disclosure surface is the default `prom-client` set only.** Verified against `prom-client@15.1.3`: `process_*`, `nodejs_heap_*`, `nodejs_eventloop_lag_*`, `nodejs_active_*`, `nodejs_gc_duration_seconds`, `nodejs_version_info{version="v20.x",...}`. No DB host, DSN, or query text. Mildly sensitive at worst, and now opt-in + documented — not a blocker.
- **Exposition format parses.** `registry.contentType` is `text/plain; version=0.0.4; charset=utf-8` (verified against the installed package), so Prometheus can scrape it.
- **Instances created once.** `createMetrics()` builds a private `Registry` (not the global default), and `buildPlugins()` runs exactly once at boot (`src/index.ts:12`). No duplicate-registration throw, no per-request allocation beyond a `WeakMap` set. 11 histogram buckets × bounded labels is fine.
- **#187 (readiness):** no path collision — it serves `/readiness`, which is already in `KNOWN_ROUTES`.

### Non-blocking nits

**1. `http_requests_in_flight` goes permanently negative (worth fixing before merge).**

Yoga registers its built-ins *before* user plugins (`graphql-yoga/cjs/server.js:166` `useHealthCheck`, `:171` `useCORS`, `:215` `...options.plugins`), and `@whatwg-node/server`'s `stopEarly()` only stops the `onRequest` chain — **all `onResponse` hooks still run** (`createServerAdapter.js:55-95`). So `/healthcheck` and every CORS `OPTIONS` preflight short-circuit before `metrics.onRequest` (no `inc`) but still reach `metrics.onResponse` (`dec`). Both happen constantly in production: k8s liveness on `/healthcheck`, and `mina-explorer` sends `Content-Type: application/json`, which is not CORS-safelisted, so browsers preflight.

I reproduced it against real Yoga 4.0.4 with your plugin's exact logic:

```
events: [["dec","/healthcheck","GET",200],["dec","/","OPTIONS",204],
         ["inc","/","POST"],["dec","/","POST",200], ...]
FINAL inFlight = -2
```

It only ever drifts down, so the gauge is unusable after a few minutes. Same thing will happen with #185's limiter if a 429 short-circuits ahead of this plugin. Fix — gate the `dec` (and the duration observation) on having actually seen the request, keeping `requestsTotal` unconditional so short-circuited routes are still counted:

```diff
--- a/src/server/metrics.ts
+++ b/src/server/metrics.ts
@@
     onResponse({ request, response }) {
       const route = routeOf(request.url);
       if (route === METRICS_PATH) return;
 
-      metrics.inFlight.dec();
       const labels = {
         method: request.method,
         route,
         status: String(response.status),
       };
       metrics.requestsTotal.inc(labels);
-      const start = startTimes.get(request);
-      if (start !== undefined) {
-        metrics.requestDuration.observe(labels, (Date.now() - start) / 1000);
-      }
+
+      // Yoga's built-in plugins (health check, CORS preflight) and any plugin
+      // ordered ahead of this one can end the response before our `onRequest`
+      // runs — `onResponse` still fires. Only balance the gauge for requests we
+      // actually saw on the way in, otherwise it drifts negative.
+      const start = startTimes.get(request);
+      if (start === undefined) return;
+      startTimes.delete(request);
+      metrics.inFlight.dec();
+      metrics.requestDuration.observe(labels, (Date.now() - start) / 1000);
     },
```

Verified with the patched logic against the same server: `counts: [GET /healthcheck 200 →1, OPTIONS / 204 →1, POST / 200 →1]`, `durations observed: 1`, `FINAL inFlight = 0`. (Tradeoff worth a doc line: short-circuited routes appear in `http_requests_total` but not in `http_request_duration_seconds_count`.)

Regression test — this fails with `-2` on the current head:

```ts
test('in-flight gauge is balanced for short-circuited requests', async () => {
  const metrics = createMetrics({ collectDefault: false });
  const yoga = createYoga({
    schema,
    graphqlEndpoint: '/',
    healthCheckEndpoint: '/healthcheck',
    cors: { origin: '*', methods: ['GET', 'POST'] },
    plugins: [useMetrics(metrics)],
  });

  await yoga.fetch('http://localhost/healthcheck');
  await yoga.fetch('http://localhost/', {
    method: 'OPTIONS',
    headers: {
      origin: 'https://explorer.test',
    },
  });
  await graphql(yoga);

  const body = await (await yoga.fetch('http://localhost/metrics')).text();
  const line = body.split('\n').find((l) => l.startsWith('http_requests_in_flight '));
  assert.strictEqual(line, 'http_requests_in_flight 0');
});
```

**2. `/metrics` gets CORS headers (relevant to #184).** Because the scrape response is produced via `endResponse` in `onRequest`, Yoga's `useCORS` `onResponse` still decorates it — I measured `GET /metrics` with `Origin: https://explorer.test` returning `Access-Control-Allow-Origin: https://explorer.test`. Harmless with `CORS_ORIGIN=*`, but it means a browser on any origin can read the scrape body once `ENABLE_METRICS=true`. If #184 lands an allowlist, consider stripping CORS headers on `METRICS_PATH`, and either way keep the ingress rule as the real control.

**3. Optional: normalise `method` too.** Worst case today is 35 methods × 5 routes × ~8 statuses × 14 series ≈ 20k series from a trivial method sweep. Not a DoS, but `const KNOWN_METHODS = new Set(['GET','POST','OPTIONS','HEAD'])` with an `other` fallback caps it at ~2k for free.

**4. Rebase needed.** `mergeable: false / dirty` — `package.json` + `package-lock.json` conflict (branch bumps the lock `0.0.5 → 0.0.6`; `main` is at `0.0.9` after `3d254f7 chore: release 0.0.9`). Regenerate the lock on top of `main` rather than hand-resolving the version. The `keywords` prettier reformat is unrelated churn that widens the conflict; dropping it would make the rebase a lock-only resolution.

SanabriaRusso (dismissed):

SanabriaRusso (approved):
Approving. The round-1 must-fix is genuinely fixed, and I verified it empirically rather than by reading.

**Verification method.** I built a standalone harness running the *real* installed `graphql-yoga@4.0.4` and `prom-client@15.1.3` with the pre-fix and post-fix plugin bodies transliterated verbatim, against a live `createYoga({ healthCheckEndpoint: '/healthcheck', cors: {...} })` instance, and fired seven request shapes: normal POST, `GET /healthcheck`, `OPTIONS` preflight, throwing resolver, 404, `GET` with no query, and a 415 bad content-type.

| plugin | final `http_requests_in_flight` |
|---|---|
| pre-fix (round 1 code) | **-2** (reproduces the original report exactly: -1 after healthcheck, -2 after preflight) |
| `d9d4d6fa` | **0** — and 0 after *every* individual step |

The mechanism checks out too, not just the symptom:

- `@whatwg-node/server/esm/createServerAdapter.js:52-83` — `endResponse()` calls `stopEarly()`, which breaks only the `onRequestHooks` loop; `handleResponse()` then runs **all** `onResponseHooks` unconditionally.
- `graphql-yoga/esm/server.js:161-167` — `useHealthCheck` and `useCORS` are registered **before** `...(options?.plugins ?? [])`.

So the `startTimes` WeakMap guard at `src/server/metrics.ts:96-97` is exactly the right shape: no `inc` ⇒ no `dec`.

**No positive leak either.** The `dec()` is finally-equivalent because yoga's built-in `useErrorHandling` (`@whatwg-node/server/esm/plugins/useErrorHandling.js:29-46`) wraps `requestHandler` in try/catch *ahead of* user plugins, so the promise always resolves to a `Response` and `onResponse` always fires. The throwing-resolver / 404 / 415 cases all ended at 0. `startTimes.delete(request)` additionally makes a double-`onResponse` idempotent.

**Regression test is real.** `tests/unit/metrics.test.ts:110-133` fires healthcheck + preflight + a normal POST and asserts the exact line `http_requests_in_flight 0` (I confirmed prom-client 15.1.3 renders a zero gauge as exactly that string, so the `strictEqual` isn't brittle against a formatting difference). The `ENABLE_METRICS` gate has its own test at `:41-70`.

**Round-1 "verified OK" items all survived the rebase:** only `onRequest`/`onResponse` hooks (no `execute`/`validate` wrapping ⇒ error text unchanged); `routeOf()` buckets any unknown path to `other`, so there is no raw-URL cardinality bomb; `/metrics` stays behind `ENABLE_METRICS`; docs at `docs/getting-started.md:214` correctly state it is unauthenticated and must be restricted at the ingress.

**Non-blocking nits**

1. `metrics.ts:94` — `requestsTotal.inc(labels)` runs *before* the `start === undefined` guard, so short-circuited requests (healthcheck, preflight) land in `http_requests_total` but never in the histogram. `rate(http_requests_total)` and `rate(http_request_duration_seconds_count)` will diverge on those routes. Defensible — duration is genuinely unknowable there — just worth a comment so a future dashboard author isn't confused.

2. Narrow, pre-existing edge case: if a *user* plugin registered after `useMetrics` throws out of its own `onRequest`, `createServerAdapter.js:119-125` catches at the outer `.catch` and skips `handleResponse` entirely, so `onResponse` never runs and the gauge leaks +1. Not introduced here; only reachable if #185's or #187's hook throws synchronously.

3. Cross-PR, for **#185** rather than this PR: its `EXEMPT_PATHS` is `{/healthcheck, /readiness}` and omits `/metrics`. Both merge orders are safe for the gauge (verified: limiter-first ⇒ no `inc` and `onResponse` returns early; metrics-first ⇒ a matched `inc`/`dec` pair), but if the limiter ends up first, scrapes consume the per-IP budget. Adding `/metrics` to `EXEMPT_PATHS` is the tidy fix.

4. Cross-PR, for **#196**: its k8s path correctly uses `prometheus.io/*` pod annotations (scrapes the pod IP, not the Service), but its Compose example publishes `8080:8080` with `ENABLE_METRICS: 'true'`, which makes `/metrics` publicly reachable on a `0.0.0.0`-bound host.

**Merge note.** `src/server/plugins.ts` conflicts textually with #185 and #190 — all three insert at index 0 of `buildPlugins`. There is no *semantic* conflict (all three plugins are `onRequest`/`onResponse`-only, and this PR's WeakMap guard makes it order-independent by construction), but whoever resolves must keep #190's rewritten `useLogger` `logFn` **and** this PR's `ENABLE_METRICS` gate.

**Downstream:** no impact on mina-explorer or mina-explorer-api. No `execute`/`validate` hooks means error text and HTTP statuses are byte-identical; the only new route is off by default and neither consumer calls it.

PR #199, opened 2026-06-29
Test coverage: gate + nightly live tests for devnet/mainnet/mesa (#181)

### LINKED ISSUE #181: Test coverage: audit + add live integration tests for devnet / mainnet / mesa
## Current state (audit)
Test tiers today:
- **`tests/unit/`** — pure logic (consensus, blockchain-utils, error-handling, network-service). No DB. Fast.
- **`tests/services/*`** — service-level tests against mocked SQL (`tests/__mocks__/*.json`).
- **`tests/integration/`** — runs against a **static archive DB dump** (`fixtures/archive_db.sql`) + a consensus-vs-chain test.
- **`tests/devnet-dump/`** — downloads a **real devnet hourly dump from GCS** and exercises services against it (old + new blocks). This is our only "real production-shaped data" coverage.
- **`tests/live-network/`** — spins up a **lightnet** container, produces blocks, queries the live archive DB. Not in routine CI (~3–5 min startup).
- **`tests/live-api/`** — hits a deployed endpoint (`http://archive-node-api.gcp.o1test.net/`), **skipped by default** (`LIVE_API_TESTS=true`). `actions.test.ts` is a **placeholder**.

## Gaps
1. **No live integration tests against devnet / mainnet / mesa.** devnet exists only as a static dump; mainnet coverage is a single o1test endpoint that's skipped by default; **mesa has zero coverage**.
2. **`live-api/actions.test.ts` is an unfinished placeholder** — needs a real ZkApp-with-actions fixture (the integration dump only contains *failed* zkApp commands, so successful-command query shapes are untested — see PR #162 review notes).
3. **No coverage gate in CI.** A coverage report exists (`c8`/lcov) but no threshold is enforced.
4. **Untested/under-tested query shapes**: `zkappCommands` against real data, `events`/`actions` pending-vs-canonical on mainnet-shaped data, very large block ranges, and DB-failure/timeout error paths.
5. **No load/soak or chaos test** (DB failover, replica drop) in CI beyond the smoke-load test.

## Proposal
- Add **live integration tests against devnet, mainnet, and mesa** archive DBs (read-only), gated/scheduled (nightly) so they don't block PRs but do catch schema/query drift against each network's real data.
- Build real fixtures for `actions` and successful `zkappCommands` (resolve the placeholder).
- Enforce a **coverage threshold** in CI on `src/**`.
- Add error-path tests (DB timeout, unreachable DB → masked error).
- Consider a scheduled soak / replica-failover test.

## Acceptance
- Nightly job runs read-only integration tests against devnet + mainnet + mesa and fails on schema/query drift.
- `actions` + successful-`zkappCommands` fixtures exist and are asserted.
- CI enforces a coverage threshold.

---
Part of #163.

dkijania:
## What & why

Part of the production-readiness epic (#163). Closes #181.

Two gaps from the test-coverage audit: **no coverage threshold** in CI, and **no live integration tests against real per-network data** (devnet existed only as a static dump; mainnet was a single skipped endpoint; mesa had none).

## Changes

- **Coverage gate** — `test:coverage` now runs c8 with `--check-coverage --lines 50 --functions 50 --statements 50 --branches 80`. Current coverage (~53% lines, 95% branches) clears it; the floor blocks regressions and can be raised as the suite grows (this epic adds ~10 test files).
- **`test:live-api`** script — runs the live-api suite against `STAGING_GRAPHQL_ENDPOINT`.
- **`Live Integration` workflow** — nightly + manual; runs live-api against **devnet/mainnet/mesa** using per-network repo variables (`DEVNET_/MAINNET_/MESA_ARCHIVE_API_URL`). A network with no configured endpoint is **skipped**, so the job is green until the maintainer sets the URLs; `fail-fast: false` keeps networks independent.
- **Docs** — live-api README documents the setup and the remaining fixture work.

## Honest remaining work (documented, not silently skipped)

- The maintainer must set the `*_ARCHIVE_API_URL` repo variables for the nightly runs to actually exercise each network.
- `actions.test.ts` is still a placeholder and a successful-`zkappCommands` fixture is still needed — both require known mainnet/mesa zkApps to snapshot against. Noted in the README.

## Testing

- `npm run test:coverage` → passes the gate locally (exit 0)
- Unit suite green


SanabriaRusso:
Thanks for pushing coverage forward here — the design is solid: scheduled-only (never on PRs), `fail-fast: false`, endpoints via repo `vars` with skip-until-set, and read-only. I ran the new gate locally and it clears comfortably (stmts 53.3 / funcs 53.2 / lines 53.3 / branch 95.5 vs the 50/50/50/80 floors, exit 0), so it won't block the production-readiness train.

One heads-up before the `*_ARCHIVE_API_URL` variables get set: `test:live-api` runs the entire `tests/live-api/*` suite against all three networks, and the two `events.test.ts` "Canonical Chain" cases assert `deepStrictEqual` against network-specific snapshots — `fixtures/…_65_66.json` (address `B62qp…J1fu`, blocks 433465–433466) and `…_84_83.json` (429984–439983). Those fixtures capture one network's `stateHash` / `parentHash` / `eventData`, so on the other two networks the same block heights return different data and the assertions go **deterministically red** — which would look like drift when it isn't.

Since the goal is catching *real* per-network drift, it'd help to gate the fixture-pinned `events` cases to their origin network while running network-agnostic shape/relationship checks (the `blocks.test.ts` style) everywhere, e.g. pass the matrix network into the step:

```yaml
env:
  STAGING_GRAPHQL_ENDPOINT: ${{ steps.endpoint.outputs.url }}
  LIVE_API_NETWORK: ${{ matrix.network }}
```
```ts
// events.test.ts
const FIXTURE_NETWORK = 'mainnet'; // wherever the J1fu fixtures were captured
const canonical = process.env.LIVE_API_NETWORK === FIXTURE_NETWORK ? describe : describe.skip;
canonical('Canonical Chain', () => { /* the deepStrictEqual cases */ });
```

To be clear, `networkState.test.ts`'s `pendingMax − canonicalMax === 290` is **not** network-specific and is fine as-is — that's the protocol-wide finality depth `k = 290` (the test documents it exactly that way), so it holds on every Mina network and is precisely the kind of cross-network invariant worth keeping. The only thing to gate is the fixture-pinned `events` data.

All nightly-only, so nothing blocks merges today — just flagging it so turning on the endpoints doesn't immediately light up part of the matrix. Thanks again for closing the coverage gap!

dkijania:
Thanks @SanabriaRusso — fixed in `1d6607b`, and you were right on both counts. Investigating it turned up a second, larger problem sitting underneath.

**Your finding, confirmed empirically.** Rather than assume where the J1fu fixtures came from, I queried all three live networks for the fixture's exact block range:

| network | `events(address: B62qp…J1fu, from: 433465, to: 433466)` |
|---|---|
| **mainnet** | returns the fixture's exact `stateHash` `3NLrHxJXQbR3AHRjUMfLz1iy96xt1CDouJLFarkQfUKvodybLhJ5`, `parentHash`, tx hash, and `eventData` |
| devnet | `[]` |
| mesa | `[]` |

So the fixtures are **mainnet**, and on the other two the `deepStrictEqual` compares live-empty against mainnet data — deterministically red, exactly as you said. They're now gated to `FIXTURE_NETWORK = 'mainnet'` with the matrix network passed in as `LIVE_API_NETWORK`, per your suggestion. When it's unset (local runs) they still run, so nobody pointing the suite at mainnet by hand loses coverage.

**The bigger one: the fixtures never loaded at all.** `events.test.ts` read them relative to `__dirname`, but `test:live-api` runs compiled out of `build/` and `tsc` emits only JavaScript — the `.json` files stay in the source tree. So the read threw `ENOENT` **at module load**, taking the whole file down before a single test ran:

```
Error: ENOENT: no such file or directory, open
  '…/build/tests/live-api/fixtures/B62qpHtWX41NstxzzUe8xooKogqomDwgJ4CN8J3V2274v5B9dnfJ1fu_65_66.json'
```

That's worse than the 2-of-3 you predicted — the events suite would have failed on **all three** networks, mainnet included, and as a module-load crash rather than an assertion, so the reason wouldn't have been obvious. Fixtures now resolve from the source tree via `process.cwd()`, matching the existing pattern in `tests/integration/setup.ts`.

**Verified end-to-end**, since a fix here is only worth as much as its proof:
- `LIVE_API_NETWORK=devnet` → module loads, both fixture suites report `# SKIP`, nothing red.
- `LIVE_API_NETWORK=mainnet` against a live mainnet archive endpoint → both cases **run and pass**.

**And +1 on your `k = 290` point** — agreed, `networkState.test.ts` is protocol-wide finality depth, not network-specific. Left running everywhere; that cross-network invariant is the whole point of the matrix. Only the fixture-pinned data is gated.

SanabriaRusso:
**Verdict: MERGEABLE** ✅

Second pass at `1d6607b`. Both of the maintainer's points are genuinely fixed, and the fix is verifiable from the tree rather than just asserted.

### What I checked

- **Maintainer's fixture-gating point — addressed.** `LIVE_API_NETWORK` is plumbed through at `.github/workflows/live-integration.yaml:53` and consumed by `describeOnFixtureNetwork` in `tests/live-api/events.test.ts`; both fixture-pinned `Canonical Chain` blocks are wrapped, and `LIVE_API_NETWORK` unset still runs them (local behaviour preserved). `networkState.test.ts` correctly left ungated.
- **The ENOENT-at-module-load claim is real, not a story.** `tsconfig.json` has `include: ["src/**/*","tests/**/*",…]` and `resolveJsonModule`, but the fixtures are read with `readFileSync`, never `import`ed — so `tsc` has no reason to emit them into `build/` and doesn't. The `process.cwd()` fix matches the existing precedent at `tests/integration/setup.ts:22`. Good catch.
- **The coverage gate is honest.** `--check-coverage --lines 50 --functions 50 --statements 50 --branches 80` is actually passed to `c8` (`package.json` `test:coverage`), invoked bare at `.github/workflows/unit-tests.yaml:26` — no `continue-on-error`, no `|| true`. The only `|| true` in that workflow is on the cosmetic `text-summary` step (line 30), which is fine. Verified against the real numbers from this PR's own CI run (`unit-tests`, run 29571039580): **stmts 53.34% (1100/2062), branches 95.57% (108/113), funcs 53.24% (41/77), lines 53.34%** — clears the floors, so the gate isn't set to a number it can never reach *or* to one it can never fail.
- **It does not gate the merge train.** Branch protection on `main` requires exactly `Run-Tests` and `Linting` (`/branches/main/protection`); `unit-tests` is not required, and `run-tests.sh` (which backs `Run-Tests`) never invokes `test:coverage`. So the gate is advisory for the ~20 open PRs — nobody has to add tests to merge. For the record, the headroom if it were ever promoted to required: ~69 uncovered statements out of 2062, so a PR adding a single ~70-line untested module would trip it. Worth knowing before anyone makes it required.
- **`src/db/sql/` and `src/services/` are in scope.** `--include 'build/src/**'` does cover them — `src/db/sql/events-actions/queries.ts` (8.16%), and all five `src/services/*` modules appear in the report. The include is not quietly omitting the modules that matter. (One real gap below.)
- **Live tests are scheduled-only and cannot go permanently red for missing credentials.** `on: schedule` + `workflow_dispatch`, no `pull_request` — the merge train is never gated on external infra. No `secrets.*` at all; endpoints come from `vars.*_ARCHIVE_API_URL` with a skip-until-set branch, so an unconfigured network is green rather than a red job nobody can fix.
- **Endpoint load is a non-issue, with numbers.** The whole suite is **9 `client.request` calls per network per night** (5 blocks, 3 events of which 2 are mainnet-gated, 1 networkState), max `limit: 100`, one events query over a 10k height range. Nothing resembling `mina-explorer`'s `ANALYTICS_BLOCK_LIMIT = 2000`. Against #185's default `RATE_LIMIT_MAX=600 / RATE_LIMIT_WINDOW_MS=60000` per IP, ~9 requests/night from one runner IP has no interaction whatsoever — no coordination needed between these two PRs.
- **Downstream: zero surface.** No `src/` change, no schema change (`GraphQL Inspector` green), so nothing touches error-message text, CORS, or query shapes for `mina-explorer` / `mina-explorer-api`.

### Non-blocking nits

**1. No job timeout — this is the one I'd actually fix before turning the endpoints on.** `test:live-api` sets no `--test-timeout` (unlike `test:live-network`'s `600000` and `test:devnet-dump`'s `900000`), Node's default `--test-timeout` is `Infinity`, and `graphql-request`/undici applies no overall request deadline. A wedged endpoint therefore hangs the run until GitHub's **6-hour** default job timeout, on up to three matrix legs. Two-line fix:

```yaml
  live-api:
    name: live-api (${{ matrix.network }})
    runs-on: ubuntu-latest
    timeout-minutes: 15
```
```json
"test:live-api": "npm run build && node --enable-source-maps --test-timeout 600000 --test ./build/tests/live-api/*.test.js",
```
(10 min per-test under a 15 min job budget keeps the two consistent, matching the `live-network` convention.)

**2. Nightly failures notify nobody reliably.** GitHub emails only the account that last touched the cron, and scheduled-workflow email is easy to filter away — the classic way a nightly goes red for a month. Cheapest durable option, appended to the job:

```yaml
      - name: Open an issue on failure
        if: failure() && github.event_name == 'schedule'
        uses: actions/github-script@v7
        with:
          script: |
            await github.rest.issues.create({
              owner: context.repo.owner,
              repo: context.repo.repo,
              title: `Nightly live-api failed: ${{ matrix.network }}`,
              body: `${context.serverUrl}/${context.repo.owner}/${context.repo.repo}/actions/runs/${context.runId}`,
              labels: ['live-integration'],
            });
```
This needs `permissions: { contents: read, issues: write }` at the workflow level.

**3. `--all` gap: `plugins.ts` and `archive-node-adapter.ts` are outside the denominator.** `c8`'s `--all` defaults to `false`, so `--include 'build/src/**'` only measures files the unit tests actually load. Absent from the report entirely: `src/server/plugins.ts` (where masking/CORS/#185's rate limiter live), `src/db/archive-node-adapter/archive-node-adapter.ts` (all DB access), `src/context.ts`, `src/index.ts`, `src/tracing/jaeger-*.ts`. The 53% is therefore "53% of the loaded subset", and adding an untested module *raises* the reported number by not appearing. Adding `--all` (and excluding the generated types, which would otherwise swamp it) is the honest form:

```
--all --exclude 'build/src/resolvers-types.js'
```
Deliberately **not** a blocker: flipping that today drops the real number below 50 and the floors would have to be re-set in the same commit. Better as a follow-up once this epic's ~10 test files land — just worth writing down so the 50 isn't mistaken for whole-`src` coverage later.

**4. Two live assertions pass vacuously on an empty result set.** `tests/live-api/blocks.test.ts:97` and `:111` are bare `for` loops with no length precondition, and `:135` likewise — against a network whose archive returns `[]` they're green while asserting nothing. Pre-existing, but this PR is what starts running them nightly against three networks, so it's the moment to add:

```ts
assert(data.blocks.length > 1, 'need at least 2 blocks to check ordering');
```

**5. `networkState`'s strict `=== K`.** Agreed with @SanabriaRusso that `k = 290` is protocol-wide, not network-specific — no argument there. The narrower point is that `pendingMax - canonicalMax === 290` also encodes *"canonicalization is fully caught up"*, which is an operational property of the archive, not of the protocol. `src/services/network-service/network-service.ts:54` returns `canonicalMaxBlockHeight = 0` when the canonical row is missing, so on a network younger than 290 blocks or one whose canonicalization is lagging, the diff is `pendingMax`, not `290` — plausible on a fresh mesa RC. `>=` keeps the invariant (canonical can never be closer than `k` to the tip) without the false positive:

```ts
assert(
  pendingMaxBlockHeight - canonicalMaxBlockHeight >= K,
  `canonical tip should trail pending by at least k=${K}`
);
```

**6. Two small housekeeping items.** `live-integration.yaml` pins `node-version: '22'` while `unit-tests.yaml` uses `'20'` — worth aligning so the nightly isn't the only thing exercising a different runtime. And the README table names the variables but not the values; `mesa` is ambiguous between the two the explorer ships (`mina-explorer` `src/config/networks.ts:47` `https://archive-node-api.mesa-rc.minaprotocol.com` vs `:38` `https://pre-mesa-archive-node-api.gcp.o1test.net`). Recording the intended URLs next to the table would stop the wrong one being set.

SanabriaRusso (approved):

PR #202, opened 2026-07-17
fix(scripts): detect docker compose v2 in download_db.sh

dkijania:
## What & why

`scripts/download_db.sh` hardcodes `docker-compose stop postgres` — the Compose **v1** binary, EOL since 2023 and absent on v2-only hosts. But `docs/getting-started.md` (Path C) tells users to run `docker compose up` (**v2**). So on a v2-only machine the documented setup works right up until `download_db.sh`, which then fails with `docker-compose: command not found`.

This detects whichever Compose is installed — preferring v2 — and errors clearly if neither is present.

## Provenance

This is the one salvageable hunk from #125 (Staketab). Per the review there, its other two files (`.env.example.compose`, `scripts/generate_libp2p.sh`) now conflict with #158's mainnet-3.3.1 refresh and are best dropped, so this lands the real improvement on a clean branch off `main`. Closes #125's useful part; recommend closing that PR once this merges.

## Testing

- `shellcheck scripts/download_db.sh` — clean
- `bash -n` — clean
- Simulated a v2-only host (docker present, `docker-compose` absent): detection resolves to `docker compose` and reaches `stop postgres`. Simulated a v1-only host: resolves to `docker-compose`. Missing-both: exits 1 with a clear message.

https://claude.ai/code/session_01ACRnT8ZHiAqEhusugky6fu

### DIFF
--- a/scripts/download_db.sh
+++ b/scripts/download_db.sh
@@ -16,8 +16,21 @@ DB_DIR="db"
 PG_DUMP="archive.sql"
 BASE_URL="https://storage.googleapis.com/mina-archive-dumps"
 
-# Stop docker containers 
-docker-compose stop postgres
+# Pick whichever Compose is installed. The docs tell users to run
+# `docker compose up` (Compose v2), but this script hardcoded the v1
+# `docker-compose` binary, which is EOL since 2023 and absent on v2-only hosts —
+# so the documented setup failed here with "command not found".
+if command -v docker &> /dev/null && docker compose version &> /dev/null; then
+  DOCKER_COMPOSE=(docker compose)
+elif command -v docker-compose &> /dev/null; then
+  DOCKER_COMPOSE=(docker-compose)
+else
+  echo "Error: neither 'docker compose' (v2) nor 'docker-compose' (v1) found" >&2
+  exit 1
+fi
+
+# Stop docker containers
+"${DOCKER_COMPOSE[@]}" stop postgres
 
 # clear db and data directories
 rm -rf "$DB_DIR"


SanabriaRusso:
**Verdict: MERGEABLE** ✅

Dev-tooling only — `scripts/download_db.sh` is never in the request path, so no impact on `mina-explorer` or `mina-explorer-api` (no schema, resolver, CORS, or error-text surface touched).

## What I checked
- **Detection shape is correct.** `scripts/download_db.sh:23` tests the *exit status* of `docker compose version` (not string-matching its stdout), with `&> /dev/null` suppressing both streams. Falls back to `command -v docker-compose` at `:25`, errors at `:28`.
- **No zsh word-splitting hazard.** Shebang is `#!/usr/bin/env bash` (`:1`) and the command is stored as an array `DOCKER_COMPOSE=(docker compose)` invoked as `"${DOCKER_COMPOSE[@]}" stop postgres` (`:33`) — correct under bash regardless, and immune to the unquoted-`$VAR` trap.
- **`set -eo pipefail` (`:2`) is fine here.** Both probes run as `if`/`&&` conditions, which are exempt from `errexit`, so a missing binary selects the fallback instead of aborting. No `set -u`, and `DOCKER_COMPOSE` is assigned on every path that reaches `:33` anyway.
- **Fails loudly**: clear message to stderr + `exit 1` (`:28-29`), no silent fall-through to a confusing downstream error.
- **Migration is complete, not partial.** Grepped `docker` across the whole file at head `a1052e5` — the only invocation is `:33`; no bare `docker-compose` call survives outside the detection block and the comment text.

### Non-blocking nits
- `command -v docker` at `:23` is redundant — `docker compose version` already fails (exit 127) if `docker` is absent. Harmless.
- Branch is `BEHIND` `main`; rebase/update before merging.

SanabriaRusso (approved):

PR #225, opened 2026-08-31
Add bounded verification-key update query

### LINKED ISSUE #224: Add a bounded verification-key update query for zkApp discovery
## Problem

The API currently supports address-oriented events and actions queries, while the blocks query does not expose zkApp account updates. As a result, a client that knows a verification-key hash cannot discover the accounts whose applied updates set that verification key without connecting directly to the archive PostgreSQL database.

This is useful for generic zkApp discovery and indexer workflows. MinaGuard is one motivating consumer, but the capability should remain application-independent.

## MinaGuard use case

[MinaGuard](https://github.com/zksecurity/mina-guard) is a web and desktop multisig wallet implemented as a Mina zkApp. Its full-mode indexer must automatically discover deployed MinaGuard vaults, including vaults that are not already present in the user's browser storage or supplied manually by address. After discovery, MinaGuard reads each contract's configuration and events to determine which vaults are associated with the connected wallet and to index their proposals.

A deployed contract is identified by the MinaGuard verification-key hash. The existing address-oriented event and action queries cannot bootstrap this process: they require the contract address, but discovering those addresses is the operation MinaGuard needs to perform. Fee-payer or deployer addresses are also not a reliable substitute for contract identity, and the existing blocks query does not expose the account updates that set verification keys.

MinaGuard currently needs direct, read-only access to the archive PostgreSQL database to perform full discovery. That couples the application to the archive schema, requires distributing database credentials to the backend, and prevents an API-only deployment. Without the database path, MinaGuard can only provide a lite mode based on already-known or manually supplied contract addresses, which can omit vaults users expect to see.

The proposed bounded query lets the MinaGuard indexer scan forward incrementally by block range for its known verification-key hash, persist a cursor, account for canonical and pending chain state, and then use the existing address-oriented APIs once contract addresses are known. This preserves full discovery without exposing PostgreSQL and keeps the API primitive generic for other zkApp indexers.

## Proposed API

```graphql
input VerificationKeyUpdateFilterInput {
  verificationKeyHash: String!
  from: Int!
  to: Int!
  status: BlockStatusFilter
}

type VerificationKeyUpdate {
  accountUpdateId: String!
  address: String!
  tokenId: String!
  verificationKeyHash: String!
  blockInfo: BlockInfo!
  transactionInfo: TransactionInfo!
}

type Query {
  # existing fields...
  verificationKeyUpdates(input: VerificationKeyUpdateFilterInput!): [VerificationKeyUpdate!]!
}
```

## Proposed semantics

- `from` is inclusive and `to` is exclusive.
- Both bounds are required; `to` must be greater than `from`.
- The interval is capped by the configured block-range limit.
- Only applied zkApp commands are returned.
- Orphaned blocks are excluded by default; callers may use the existing block-status filter for canonical or pending results.
- Results represent individual verification-key-setting account-update occurrences and use a deterministic order.
- The query returns the account address and token ID as well as existing block and transaction metadata.

## Implementation notes

The archive schema already links:

`zkapp_account_update_body -> zkapp_updates -> zkapp_verification_keys -> zkapp_verification_key_hashes`

and account updates to commands and blocks. The SQL should start from the required verification-key hash and bounded block range, and the relevant columns should be indexed or covered by existing indexes. Tests should cover validation, filtering, chain status, deterministic ordering, schema filtering through `ENABLED_QUERIES`, and the static archive dump.

I plan to follow this issue immediately with an implementation PR.

mellowcroc:
Closes #224.

## Summary

Add a generic `verificationKeyUpdates` GraphQL query so clients can discover zkApp accounts by the verification key set in an applied account update, without direct archive PostgreSQL access.

The query returns the account address and token ID together with the existing block and transaction metadata shapes.

## Query contract

- requires an exact `verificationKeyHash`
- requires an inclusive `from` and exclusive `to` block range
- rejects empty, reversed, or over-limit ranges with `BLOCK_RANGE_ERROR`
- returns only applied zkApp commands on the canonical + best pending chain
- supports the existing `ALL`, `CANONICAL`, and `PENDING` status filter
- orders occurrences deterministically by block, transaction, and account-update position
- participates automatically in `ENABLED_QUERIES` schema filtering

## Implementation

The SQL starts from the bounded best-chain block set, uses `UNNEST ... WITH ORDINALITY` for account-update IDs, and follows the archive relation from `zkapp_account_update_body.update_id` through `zkapp_updates` to the verification-key hash. This distinguishes an account update that sets a verification key from one that merely references a verification-key precondition.

The schema-generated resolver types were regenerated. Shared block and transaction metadata mappers now accept the exact row fields they consume, allowing the new service to reuse them without casts to an event row.

## Validation

- `npm run build`
- `npm run lint`
- `npm run test:unit`
- `npm run test:integration` (63 tests passed, including a static-dump check that failed deployments are excluded)


schema.graphql
```diff
@@ -216,6 +238,9 @@ type Block {
 type Query {
   events(input: EventFilterOptionsInput!): [EventOutput]!
   actions(input: ActionFilterOptionsInput!): [ActionOutput]!
+  verificationKeyUpdates(
```
dkijania, line 241: @SanabriaRusso 

Flagging a policy question for us, not a change request for you — no action needed on your side @mellowcroc 

`docs/versioning.md` says changes that alter "the set of exposed queries" ship  "disabled by default behind an environment flag". `ENABLED_QUERIES` is an allowlist only when it is set, so by default this new query is exposed. Given the cost numbers in my comment on the SQL, we may decide this ships default-off at first. 

Separately: a new query is a MINOR change under that same policy, so `package.json` needs `1.1.0`. We can fold that into the release, so leave it unless we ask.

src/db/sql/verification-key-updates/queries.ts
```diff
@@ -0,0 +1,116 @@
...
+  let statusClause = "AND b.chain_status <> 'orphaned'";
+
+  if (status !== BlockStatusFilter.all) {
+    params.push(status.toLowerCase());
+    statusClause = `AND b.chain_status = $${params.length}`;
+  }
+
+  return client.unsafe(
+    `
+      WITH RECURSIVE pending_chain AS (
```
dkijania, line 40: No change needed from you — noting this so it is on record.

This is the third implementation of the pending-chain walk in the repo. The other two are `fullChainCTE` and `getZkappsWithPendingEventsQuery` in `src/db/sql/events-actions/queries.ts`. This one differs in two places: it seeds from `MAX(height) WHERE chain_status = 'pending'` rather than `MAX(height)`, and it puts the `chain_status <> 'canonical'` guard on the parent rather than the child.

Both variants behave the same on normal archive data, so this is not a defect. We recently shipped #209 to fix a chain-position defect in one of the copies, so we want one shared helper - but that touches SQL currently in our merge train and would conflict with this PR. We will do it separately, after this lands. Leave yours as is.

src/db/sql/events-actions/queries.ts
```diff
@@ -553,6 +553,7 @@ export const USED_TABLES = [
   'zkapp_field',
   'zkapp_verification_key_hashes',
   'zkapp_verification_keys',
+  'zkapp_updates',
```
dkijania, line 558: Adding `zkapp_updates` is correct. The query also reads `public_keys` and `tokens`, and neither is in the list. That gap is pre-existing -  `accountIdentifierCTE` already uses both - but `checkSQLSchema()` is a startup  guard, so it is worth closing now that a second query depends on them.

src/db/sql/verification-key-updates/queries.ts
```diff
@@ -0,0 +1,116 @@
...
+) {
+  const params: (string | number)[] = [verificationKeyHash, from, to];
+  let statusClause = "AND b.chain_status <> 'orphaned'";
+
+  if (status !== BlockStatusFilter.all) {
+    params.push(status.toLowerCase());
+    statusClause = `AND b.chain_status = $${params.length}`;
+  }
+
+  return client.unsafe(
```
dkijania: This is safe.  `verificationKeyHash`, `from`,  `to` and the lower-cased status all go through `params`, and the only interpolated value is `params.length`, an integer. No injection path.

But it is the only raw-string query in `src/db/sql/`, and `postgres.js` `unsafe()` does not use prepared statements, so the query is re-planned on every request — I measured 20.7 ms of planning on a wide range. The conditional fragment pattern from `blocksAccessedCTE` keeps the prepared statement:
  ```
      ${status === BlockStatusFilter.all
        ? client``
        : client`AND b.chain_status = ${status.toLowerCase()}`}
 ```
I deliberately left `unsafe` in place in my PR so your unit test keeps passing
unchanged, and left this conversion to you.

tests/services/verification-key-updates-service/verification-key-updates-service.test.ts
```diff
@@ -0,0 +1,144 @@
...
+          hash: '5Jtransaction',
+          memo: 'memo',
+          authorizationKind: 'Proof',
+          sequenceNumber: 3,
+          zkappAccountUpdateIds: [41, 42],
+        },
+      },
+    ]);
+    assert.deepStrictEqual(calls[0].params, ['123456789', 100, 110]);
+    assert.match(calls[0].query, /bzc\.status = 'applied'/);
```
dkijania: These match the SQL *text*, not its behaviour, and will keep passing through any change that preserves the substring. The range-validation tests below them are the right thing for a unit test to cover here. I would just drop these two, and they have to go anyway when you make change  from `src/db/sql/verification-key-updates/queries.ts`

src/db/sql/verification-key-updates/queries.ts
dkijania: I tried this before suggesting it, because it is not the one-line change it looks
  like. It touches three files:

  *`queries.ts`* — the `params` array and the `$N` arithmetic go away:
                                                    
      const statusClause =
        status === BlockStatusFilter.all
          ? client`AND b.chain_status <> 'orphaned'`
          : client`AND b.chain_status = ${status.toLowerCase()}`;
  
      return client<VerificationKeyUpdateDatabaseRow[]>`
        ...
  
  Then `$1` -> `${verificationKeyHash}`, `$2` -> `${from}`, `$3` -> `${to}`, and `${statusClause}` where it was. One substitution is not obvious: in the SELECT  list the cast has to stay, as `${verificationKeyHash}::text AS verification_key_hash`, or Postgres cannot infer the parameter type in that  position.
  
  *`verification-key-updates-service.ts`* — a free cleanup falls out. The tagged template is generic, so this cast is no longer needed:

      -    const updates = (rows as unknown as VerificationKeyUpdateDatabaseRow[]).map(
      +    const updates = rows.map(
  
  Watch the knock-on: the `VerificationKeyUpdateDatabaseRow` import then becomes unused and `npm run lint` fails. That one caught me.

dkijania (changes requested):
Just some nits. Also i created pr into your branch on your repo for things that are on us. I realised that after review of this PR

src/db/sql/events-actions/queries.ts
mellowcroc: done

src/db/sql/verification-key-updates/queries.ts
mellowcroc: done

tests/services/verification-key-updates-service/verification-key-updates-service.test.ts
mellowcroc: removed the tests

SanabriaRusso (approved):
## Approve — backwards-compatible, and the bound is real

Independent review focused on (1) downstream backwards compatibility, (2) the new error surface,
(3) query cost / DoS, (4) SQL correctness. Everything below was **executed**, not read off the diff.
Head reviewed: `04bfb5d8`.

---

### 1. Schema change is strictly additive — verified by count, not by eye

```
git diff origin/main...04bfb5d8 -- schema.graphql        | grep -c '^-[^-]'   → 0
git diff origin/main...04bfb5d8 -- src/resolvers-types.ts | grep -c '^-[^-]'  → 0
```

Added: `input VerificationKeyUpdateFilterInput`, `type VerificationKeyUpdate`,
`Query.verificationKeyUpdates`. Removed/renamed/retyped: **nothing**. No nullability tightening on
any pre-existing field. `resolvers-types.ts` mirrors that exactly — new entries appended to
`ResolversTypes`, `ResolversParentTypes`, `QueryResolvers` and `Resolvers`, nothing dropped.

The mapper change flagged in the PR description is the one place a shape change could have hidden,
and it does not. `src/blockchain/utils.ts:4-31` changes

```ts
createBlockInfo(row: ArchiveNodeDatabaseRow)  →  createBlockInfo(row: BlockInfoRow)
```

where `BlockInfoRow = Pick<ArchiveNodeDatabaseRow, …>`. That is a **widening** of a parameter type
(contravariant): every existing caller still type-checks, and both function *bodies* are byte-identical
to `main`. The emitted JavaScript is unchanged. The existing events/actions integration tests that
exercise both mappers still pass (77/77, below).

**Downstream impact: none.** Neither `mina-explorer` nor `mina-explorer-api` references
`verificationKeyUpdates`, and no field either consumer selects today changed name, type or nullability.

### 2. `BLOCK_RANGE_ERROR` is not a new error code, and it does not open a circuit breaker

`throwBlockRangeError` (`src/errors/error.ts:34`, code `BLOCK_RANGE_ERROR`) already exists on `main`
and is shared with events/actions; this PR adds no new error string. Verified live against a running
server (`node build/src/index.js`, real Postgres):

```
POST / {verificationKeyUpdates(input:{verificationKeyHash:"x",from:10,to:10})}
→ HTTP/1.1 200 OK
  {"data":null,"errors":[{"message":"to must be greater than from",
    "extensions":{"code":"BLOCK_RANGE_ERROR","status":400}}]}

POST / {… from:0,to:20000}
→ HTTP/1.1 200 OK
  {"errors":[{"message":"The block range is too large. The maximum range is 10000", …}]}
```

**HTTP 200 with a populated `errors[]`.** So for `mina-explorer-api`: `_execute` sees a normal
GraphQL error response, not a bare 4xx, so `breaker.record_failure()` is never reached. And neither
message contains `"Cannot query field"`, `"Unknown argument"` or `"Unknown type"`, so the
capability cache in `app/upstream/graphql.py:33-42` is not poisoned. For `mina-explorer`, no
`HTTP error: NNN` throw path is entered.

### 3. Cost / DoS — measured on a real 691k-block devnet archive

I loaded the current hourly devnet dump (`devnet-archive-dump-2026-09-01_1200`, 1.4 GB SQL →
3.0 GB database): **691,103 blocks, max height 552,287, 176,348 `zkapp_updates`,
369,074 `zkapp_account_update_body`, 1,593 verification keys.** `EXPLAIN (ANALYZE, BUFFERS)` on the
exact SQL from `src/db/sql/verification-key-updates/queries.ts`, maximum 10,000-block range:

| case | rows out | execution |
|---|---|---|
| most common vk hash (564 `target_updates`), tip range 542287–552287 | 0 | **196–265 ms** |
| same hash, range 398000–408000 (a range that actually matches) | 260 | **277–457 ms** |
| **hash that matches nothing at all**, max range | 0 | **9–125 ms** |

The nonexistent-key case is the *cheapest*, not the most expensive — `target_updates` comes back
empty and the semi-join short-circuits. This is **not** the defect shape from #162: the guard runs
before any SQL (`verification-key-updates-service.ts:63`, and the unit test asserts
`calls.length === 0` on rejection), and the join is driven **from** the bounded block set
(`Index Scan using idx_blocks_height on blocks b, Index Cond: (height >= … AND height < …)`),
with the vk set applied as a semi-join predicate. Range work is genuinely bounded by `from`/`to`.

I independently reproduce the ~199 ms figure in the code comment at `queries.ts:56-62`.

**One dimension the range guard does not bound** (non-blocking, evidence below). Cost is linear in
`|target_updates|` — the number of `zkapp_updates` rows that set the requested key — and nothing
caps that. On synthetic data at realistic table ratios (3.9M `zkapp_updates`, 3M
`zkapp_account_update`) I measured, over a **48-block** range returning 8 rows:

```
target_updates       2 →     0.8 ms
                   101 →     6.4 ms
                  1001 →    21.9 ms
                 10001 →   208.8 ms
                100001 →  2230.7 ms
               1000001 → 24481.0 ms   ← planner flips to a nested loop, 10M rows discarded
```

Clean linear growth, ~24.5 µs per target row, entirely independent of the block range. **This is
not reachable on real data**: `zkapp_updates` is hash-consed by the archive node, so the observed
**maximum across all 1,593 keys and all 691k blocks of devnet history is 564** — 177× below the
100k where it would start to hurt. I am recording the shape, not asking for a change.

Two cheap, optional hardenings, both outside this PR's blast radius:
- `zkapp_updates.verification_key_id` has **no index** (`\d zkapp_updates` shows only the PK; a
  Postgres FK constraint does not create one), so `target_updates` always sequential-scans the
  table — 21.5 ms of the 265 ms above on devnet, and range-independent, so a 1-block query pays it
  too. `CREATE INDEX CONCURRENTLY idx_zkapp_updates_verification_key_id ON zkapp_updates (verification_key_id);`
  removes it. That is an archive-DB index, not a repo change — see the joint ask in §6.
- #182 (`PG_STATEMENT_TIMEOUT`) is the right backstop for the tail. Worth landing before this
  endpoint is publicly exposed; it is not a reason to hold this PR.

### 4. SQL correctness — the three claims in the description hold

- **Sets vs. precondition.** The join is on `zaub.update_id` (the key the update *writes*), not
  `zaub.verification_key_hash_id` (the key it *requires*). Confirmed by the schema — both columns
  exist on `zkapp_account_update_body` — and by the integration test
  *"a verification-key precondition is not a verification-key update"*.
- **Failed / non-applied commands excluded.** `WHERE bzc.status = 'applied'`, covered by
  *"excludes a different verification key, a failed command, a precondition-only update, and an
  orphaned block"*.
- **Ordering is total.** `UNNEST … WITH ORDINALITY` ordinality is the account-update position
  within `zc.zkapp_account_updates_ids`; `state_hash` in the sort key is what disambiguates two
  competing tips at one height. Covered by three separate tests including
  *"repeats the same order on every run"*.
- **Chain selection** matches `fullChainCTE` in `events-actions/queries.ts` in effect
  (`distance_from_max_block_height` is computed identically, `(SELECT max(height) FROM blocks) - height`).
  It differs in the pending seed, which @dkijania already reviewed and explicitly said to leave.
  One small improvement over the shared helper: the `ALL` branch adds `chain_status <> 'orphaned'`,
  which the recursive walk's `<> 'canonical'` guard alone would not exclude.

### 5. `ENABLED_QUERIES` — no change to what an existing deployment serves

`src/resolvers.ts` is untouched except for the new resolver entry, so the filter picks the query up
automatically. A deployment that sets `ENABLED_QUERIES=events,actions` today serves exactly what it
served before. A deployment that does **not** set it now additionally exposes
`verificationKeyUpdates` — which is precisely the `docs/versioning.md:50-54` point @dkijania raised,
and my §3 numbers support shipping it default-off first: the endpoint is cheap on today's data but
runs with neither a statement timeout (#182) nor a cost limit (#183) merged.

`checkSQLSchema()` now also requires `public_keys`, `tokens` and `zkapp_updates` at startup. All
three are present in the live devnet archive schema (verified), and `getTables` reads
`pg_catalog.pg_tables`, which is grant-independent — so no startup regression.

### 6. Cross-PR: draft #162 ("Add zkApp command range query")

Coordinated with the reviewer of #162; folding the joint conclusion in so both authors get one story:

- **No schema collision.** #162 adds `ZkappCommandFilterOptionsInput`, `ZkappCommandOutput`,
  `Query.zkappCommands` and friends — zero name or semantic overlap with this PR. The file-level
  conflict between the two is textual and append-only (schema Query block, `resolvers.ts`,
  `USED_TABLES`, the adapter pair, `integration.test.ts`); resolve by taking both sides.
- **Range guard: #162 should follow this PR, not the reverse.** This PR reuses `throwBlockRangeError`
  and the existing `BLOCK_RANGE_SIZE` with the same message text as events/actions. #162 introduces
  two new env vars for the same concept. To answer that reviewer's question explicitly: **this PR is
  content sharing `BLOCK_RANGE_SIZE`** — `verification-key-updates-service.ts:12` imports it
  unchanged from `server.js` and adds no knob of its own. No new env var is needed here, and none
  should be needed there.
- **Joint index ask, one change not two.** `tests/integration/fixtures/archive_db.sql` has 22
  indexes and **none** on any `zkapp_*` join column either query uses. Between the two PRs the
  useful set is `zkapp_account_update_body(update_id)`, `zkapp_account_update_body(account_identifier_id)`,
  `zkapp_account_update(body_id)` and `zkapp_updates(verification_key_id)`. That belongs in
  `docs/runbook.md` as an operator prerequisite, attributed to both PRs — not in either PR.
- **On whether the 199 ms was measured with a large target set:** it was not, and neither was mine —
  because a large one does not exist. Real devnet maximum is 564 (§3). The linear-cost shape is
  real; the cardinality needed to exploit it is not present in the data.

### 7. @dkijania's open items

| item | status on `04bfb5d8` |
|---|---|
| `USED_TABLES` missing `public_keys` / `tokens` | **fixed** — both added |
| `unsafe()` → tagged template (re-planning, 20.7 ms) | **fixed** — no `unsafe`, no `params` array, `statusClause` fragment as suggested |
| SQL-text-matching unit tests | **fixed** — removed; the two remaining tests assert guard-before-query behaviour |
| third pending-chain copy | explicitly "leave as is"; shared helper deferred |
| `docs/versioning.md` default-off + `1.1.0` | policy, on the maintainers — see §5, my numbers support default-off |

Those are the author's and @dkijania's to close out; nothing there blocks from my side.

### Verified by

`npm ci && npm run build && npm run lint` (clean) · `npm run test:unit` (pass) ·
`npm run test:integration` against PostgreSQL 16 — **77/77 pass, including all 15
`verificationKeyUpdates` tests** · live server HTTP probes for the error surface ·
`EXPLAIN (ANALYZE, BUFFERS)` on a freshly downloaded 691k-block devnet archive dump ·
a synthetic scaling curve at realistic table ratios.

Nothing here is a backwards-compatibility break, a security hole, or a data-correctness bug.
Approving.

dkijania (approved):

dkijania:
@mellowcroc Approved! Thanks for your contribution. We just need to solve issues with CI. I think that this is connected with running against forked repo. I will try to patch it in another pr and ask you to update your pr

dkijania:
updating after possible ci fix

Issue #213, opened 2026-08-24
Check Schema and the npm audit gate are not required status checks, so neither can block a merge


SanabriaRusso:
## Problem

`main`'s branch protection requires only two status checks:

```
required status checks : ["Run-Tests", "Linting"]
strict                 : true
required approvals     : 1
dismiss_stale_reviews  : true
enforce_admins         : false
```

Several CI jobs that exist specifically to *prevent* a class of breakage are therefore advisory: they go red without blocking a merge. The two that matter:

### 1. `Check Schema` (graphql-inspector) is not required

`.github/workflows/graphql-inspector.yaml` runs a breaking-change check with `approve-label: expected-breaking-change`, and the `expected-breaking-change` label already exists in the repo. But because the check is not required, a PR that removes or renames a schema field goes red and merges anyway.

This matters more than usual here because #198 introduces a versioning and schema-stability policy whose Enforcement section states that a breaking schema change "fails the build … so every breaking change is a deliberate, reviewed decision." That is currently not true. The policy leans on a gate that is not wired up.

It also matters downstream. `mina-explorer-api` classifies any response containing `Cannot query field`, `Unknown argument`, or `Unknown type` as a permanent schema error (`app/upstream/graphql.py:33-42`), which drives tier fallback **and poisons its capability cache** — so an accidental field removal does not merely fail, it fails stickily until that cache is cleared.

### 2. The new `npm audit` gate (#192) will not be required either

#192 adds `.github/workflows/security.yaml` with a real, correctly-calibrated gate (`npm audit --omit=dev --audit-level=critical`, no `|| true`, no `continue-on-error`). Verified against `main`'s lockfile it currently exits 0, so it is not a permanently-red gate. But as a non-required check it cannot block anything.

## Proposed resolution

**After #192 merges** (so the check name exists and has at least one successful run on `main`), update branch protection:

```sh
gh api -X PATCH repos/o1-labs/Archive-Node-API/branches/main/protection/required_status_checks \
  -f 'contexts[]=Run-Tests' \
  -f 'contexts[]=Linting' \
  -f 'contexts[]=Check Schema' \
  -f 'contexts[]=npm audit'
```

Confirm the exact check names first — they must match the `name:` GitHub reports, not the workflow filename:

```sh
gh pr checks <any-recent-pr> --json name,bucket
```

[...]

PR #227, opened 2026-09-02
ci: make schema and build jobs work without repository secrets (#214)

### LINKED ISSUE #214: CI leaves a live GCP registry token in .npmrc inside the Docker build context on non-version-bump builds
## Problem

`.github/workflows/build.yaml` writes a live GCP access token into `.npmrc` in the workspace root:

```yaml
# .github/workflows/build.yaml:63-66
run: |
  # Only set authentication, not the default registry
  # This allows dependencies to install from public npm
  echo "//europe-southwest1-npm.pkg.dev/o1labs-192920/euro-npm/:_authToken=$(gcloud auth print-access-token)" > .npmrc
```

The cleanup that removes it is conditional:

```yaml
# .github/workflows/build.yaml:110-113
if: always() && steps.determine_npm_version.outputs.needs_version_update == 'true'
run: |
  node -e "...restore package.json version..."
  rm -f .npmrc
```

`docker/build-push-action` runs later in the same job with `context: .`. So on any build where `needs_version_update` is **false**, `.npmrc` — containing a valid registry token — is still on disk and is uploaded into the Docker build context.

`.dockerignore` (added by #189) does not list it:

```
node_modules
build
.git
.github
.env*
db
data
...
```

## Current exposure

**There is no leak today.** The `Dockerfile` never `COPY`s `.npmrc`: its only `COPY` sources are `package*.json`, `src`, `tsconfig.json`, and `schema.graphql`, and the `package*.json` glob does not match `.npmrc`. The file enters the build *context* but never a published *layer*.

This is filed as a latent hazard, not an active incident. The failure mode is one careless line away: any future `COPY . .` or `COPY . /app` — the most natural thing to write when adding a file to the image — would bake a registry credential into a layer that is pushed to both GAR and GHCR. Defending against that is precisely what `.dockerignore` is for, and the cost is one line.

## Proposed resolution

**Primary fix** — add to `.dockerignore`:

```diff
 .env*
+.npmrc
 db
```

I raised this as a non-blocking note on #189 (which introduces `.dockerignore`). If #189 merges without it, apply it directly to `main`. Either route is fine; this issue exists so it is not lost.

**Secondary fix, recommended** — make the cleanup unconditional in `.github/workflows/build.yaml`. A credential should not survive on disk because an unrelated version-bump branch was not taken:

```diff
-        if: always() && steps.determine_npm_version.outputs.needs_version_update == 'true'
+        if: always()
         run: |
-          node -e "const fs = require('fs'); ... pkg.version = '${{ steps.determine_npm_version.outputs.original_version }}'; ..."
           rm -f .npmrc
```

Keep the `package.json` version restore on its existing condition — only the `rm -f .npmrc` needs to become unconditional. Splitting them into two steps is the cleanest way to express that.

**Also worth checking**: whether the token needs to be in a file at all. `npm` reads `NPM_CONFIG_//registry/:_authToken` from the environment, which would keep the credential out of the filesystem and out of every build context by construction. That is a larger change; the two fixes above are sufficient.

## Acceptance criteria

- [ ] `.npmrc` is listed in `.dockerignore`
- [ ] `rm -f .npmrc` runs unconditionally in `build.yaml`
- [ ] Verified: after a build where `needs_version_update == 'false'`, no `.npmrc` remains in the workspace

dkijania:
## What & why

A pull request from a fork, and one opened by Dependabot, runs with a **read-only `GITHUB_TOKEN` and no repository secrets**. Two jobs assume otherwise and fail on every such pull request. This is currently red on **#225** (fork, community contribution) and **#226** (Dependabot).

### 1. `Check Schema` — could not report

```
##[error]Resource not accessible by integration - .../checks/runs#create-a-check-run
```

`kamilkisiela/graphql-inspector@master` reports by creating a check run, which needs `checks: write`. Adding a `permissions:` block does **not** help: GitHub caps the token at read-only for these events whatever the workflow asks for.

The gate is now `npx @graphql-inspector/cli@7.0.0 diff 'git:origin/main:schema.graphql' schema.graphql`, whose **exit code is the job result** — no API write, so it works from a fork. Behaviour kept: the `expected-breaking-change` label still approves a deliberate breaking change. Side benefit: `actions/checkout@master` and `graphql-inspector@master` were unpinned; both are now pinned.

### 2. `build-and-deploy` — could not authenticate, and could not check out a fork

```
##[error]google-github-actions/auth failed with: … By default, secrets are not passed
to workflows triggered from forks, including Dependabot.
```

and, on forks only, one step earlier:

```
fatal: couldn't find remote ref feat/verification-key-updates
```

This job authenticates to Google Cloud, writes a registry token, **publishes a dev npm package** and pushes an image. None of that can or should run for an untrusted pull request — the failure is GitHub's protection working, not a bug to fix by handing over secrets. So the job is now **skipped** for fork and Dependabot pull requests instead of failing at the auth step, and the checkout resolves the head SHA through `actions/checkout` rather than `git fetch origin <head ref>`, which cannot resolve a branch living in a fork.

Coverage for those pull requests is unchanged otherwise: `Run-Tests`, `Linting`, `unit-tests`, `smoke-load`, `npm audit`, `SBOM` and `Check Schema` all still run. Both required checks (`Run-Tests`, `Linting`) are unaffected.

### 3. Registry credential in the Docker build context — closes #214

`.npmrc` holds a live GCP access token, and the step that deleted it only ran when `needs_version_update == 'true'`. `docker/build-push-action` runs later in the same job with `context: .`, so on any other build the token was uploaded into the build context. Nothing copies it into a layer today; the hazard is one `COPY . .` away.

- `rm -f .npmrc` split into its own step and made unconditional.
- `.npmrc` added to `.dockerignore`.

The `package.json` version restore keeps its original condition.

## Testing

- `actionlint` — clean on both changed workflows.
- Schema gate, run locally against this repo: unchanged schema → `[success] No changes detected`, exit **0**. With `parentHash` deleted from `BlockInfo` → `✖ Field parentHash was removed from object type BlockInfo` / `[error] Detected 1 breaking change`, exit **1**. So the gate still fails the build on a breaking change.
- `.dockerignore`, verified with a throwaway image that does `COPY . /ctx` over a workspace containing a fake `.npmrc`: with the new line the file is absent from the context (build passes); with the line removed the same build **fails** — a negative control proving the test is meaningful.

## Follow-up, not in this PR

Once this lands, `Check Schema` can pass from a fork, which unblocks making it a required status check — tracked in #213.

Note on ordering: for `pull_request` events GitHub runs the workflow files from the merge ref, so #225 and #226 need their runs re-triggered after this merges (both are behind `main` anyway).

https://claude.ai/code/session_012LtdRKTkaKRAuTkEEpgPcc

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -5,6 +5,7 @@ coverage
 .github
 .memsearch
 .env*
+.npmrc
 db
 data
 *.log
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -18,18 +18,27 @@ env:
 
 jobs:
   build-and-deploy:
+    # This job authenticates to Google Cloud, publishes a dev npm package and
+    # pushes a Docker image, so it only works where repository secrets are
+    # available. GitHub passes no secrets to a pull request from a fork or from
+    # Dependabot, so the job is skipped there instead of failing at the auth
+    # step. Fork and Dependabot pull requests are still covered by Run-Tests,
+    # Linting, unit-tests, smoke-load, npm audit and Check Schema.
+    if: >-
+      github.event_name != 'pull_request' ||
+      (github.event.pull_request.head.repo.full_name == github.repository &&
+       github.event.pull_request.user.login != 'dependabot[bot]')
     runs-on: ubuntu-latest
     steps:
       - name: Checkout code
         uses: actions/checkout@v4
         with:
           fetch-depth: 0
-
-      - name: Checkout PR branch
-        if: github.event_name == 'pull_request'
-        run: |
-          git fetch origin ${{github.event.pull_request.head.ref}}
-          git checkout ${{github.event.pull_request.head.sha}}
+          # Build the head commit of the pull request rather than the merge
+          # commit, which is what this job published before. The previous
+          # `git fetch origin <head ref>` assumed the branch lived in this
+          # repository and could not resolve a fork's branch.
+          ref: ${{ github.event_name == 'pull_request' && github.event.pull_request.head.sha || github.sha }}
 
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v3.6.1
@@ -110,7 +119,13 @@ jobs:
         if: always() && steps.determine_npm_version.outputs.needs_version_update == 'true'
         run: |
           node -e "const fs = require('fs'); const pkg = JSON.parse(fs.readFileSync('package.json', 'utf8')); pkg.version = '${{ steps.determine_npm_version.outputs.original_version }}'; fs.writeFileSync('package.json', JSON.stringify(pkg, null, 2) + '\n');"
-          rm -f .npmrc
+
+      - name: Remove registry credentials from the workspace
+        # Unconditional, unlike the version restore above: `docker/build-push-action`
+        # runs later with `context: .`, so a live registry token must never be left
+        # on disk merely because the version-bump branch was not taken (#214).
+        if: always()
+        run: rm -f .npmrc
 
       - name: Determine tags
         id: determine_tags
--- a/.github/workflows/graphql-inspector.yaml
+++ b/.github/workflows/graphql-inspector.yaml
@@ -6,6 +6,14 @@ on:
     branches:
       - main
 
+# A pull request from a fork, or one opened by Dependabot, runs with a read-only
+# GITHUB_TOKEN. The previous implementation reported through a check run, which
+# needs `checks: write`, so it failed on every such pull request with
+# "Resource not accessible by integration". The gate is now the job's own exit
+# code, which needs no write access at all.
+permissions:
+  contents: read
+
 jobs:
   test:
     name: Check Schema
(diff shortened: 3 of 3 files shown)


SanabriaRusso (approved):
Approving. This closes two of the repo-level findings from the round-4 review (§4.1 `Check Schema` blind to fork/Dependabot PRs, and #214 registry token in the Docker build context) with no change to the API surface, so nothing for mina-explorer / mina-explorer-api to adapt to.

**What I verified**

*Schema gate (`graphql-inspector.yaml`)*
- The new gate ran on this PR's own merge ref (job 100292004125) and reported `[success] No changes detected` — so the `npx` path, the `git fetch` of `origin/main`, and the `git:` loader all work under the read-only token. The PR itself doesn't touch `schema.graphql`, so that only exercises the passing path.
- Ran `@graphql-inspector/cli@7.0.0 diff 'git:origin/main:schema.graphql' <file>` locally against three variants of the current schema:
  - unchanged → `[success] No changes detected`, exit **0**
  - `parentHash` removed from `Block` and `BlockInfo` → `[error] Detected 2 breaking changes`, exit **1**
  - new optional field added → `✔ Field … was added`, `[success] No breaking changes detected`, exit **0**
  The gate fails only on breaking changes, which matches the previous action's `fail-on-breaking` default; additive changes (the shape #225 uses) still pass.
- `permissions: contents: read` is explicit; no `${{ }}` interpolation of PR-controlled data into `run:` (the label check goes through `env`), and the `expected-breaking-change` label can only be applied by someone with triage rights, so a fork author cannot self-approve a breaking change.

*Build job (`build.yaml`)*
- Triggers are `pull_request` and `push: tags: v*`. The new `if:` leaves tag builds untouched, keeps same-repo PR builds (this PR's `build-and-deploy` ran and passed, 5m42s), and skips only fork/Dependabot PRs where the job could never have succeeded.
- `ref: head.sha` for PRs builds the same commit the old `git fetch origin <ref> && git checkout <sha>` step did; for tag pushes it resolves to `github.sha`, identical to before.
- `rm -f .npmrc` is now its own `if: always()` step, so the GAR token is removed before `docker/build-push-action` on every path — including release builds, where `needs_version_update == 'false'` previously left it in the context. `.dockerignore` now excludes it as defence in depth. The Dockerfile never copies `.npmrc` and `npm ci` uses the public registry, so this cannot break the image build. No `.npmrc` is committed.

**Non-blocking**
- `npx @graphql-inspector/cli@7.0.0` pins the CLI but resolves its `^` transitive deps fresh each run. Same class of exposure as the old `@master` action (strictly better, since the top-level is pinned now); a follow-up could vendor it into `devDependencies` so the lockfile covers it.
- Branch is behind `main` (no conflicts). After merge, #225 and #226 need their runs re-triggered as the description says, and #213 (make `Check Schema` required) is now actually achievable.

PR #228, opened 2026-09-02
fix(tests): make the devnet dump suite actually run

dkijania:
## What & why

The `Nightly Devnet Dump Tests` job has been red every night. The log read:

```
# Dump loaded successfully.
not ok 1 - NetworkState (devnet dump)
  error: 'relation "blocks" does not exist'
```

That is the shape of a suite that has never run to completion. **Five defects, stacked so each hid the next.** Only the first two are visible from the CI log.

### 1. The dump loaded into a different database

The devnet dumps are `pg_dump --create` dumps. They open with:

```sql
CREATE DATABASE archive WITH TEMPLATE = template0 ...;
\connect archive
```

psql obeys the dump, not its own `-d` flag. So every table landed in `archive`, while the `devnet_dump_test` database the loader had just created stayed empty — and psql exited **0**, because nothing actually failed.

Reproduced against the real dump: 692,089 blocks in `archive`, zero tables in `devnet_dump_test`.

### 2. A failed load could not surface as itself

The loader ran psql with no `ON_ERROR_STOP`, `stdio: 'pipe'` discarding stderr, and then printed `Dump loaded successfully.` unconditionally. Any load problem — this one or a future one — arrives as a confusing `relation "blocks" does not exist` at query time instead of an error at load time.

`setupTestDatabase` now reads the dump header, uses the database the dump names when it names one, keeps the previous behaviour for a plain dump, stops on the first error, reports psql's stderr in the thrown message, and **asserts the blocks arrived** rather than trusting an exit code.

The header parse rejects any database name that is not a plain identifier, since the dump is downloaded and the name reaches `DROP DATABASE`.

### 3–5. With the data finally in place, three test bugs surfaced

| Bug | Why it never passed |
|---|---|
| `date.getFullYear() >= 2023`, in two tests | Devnet's genesis is dated **2021-09-24**, so no dump can satisfy it. The intent was "not epoch 0". |
| `JOIN zkapp_accounts za ON ai.id = za.account_identifier_id`, in three tests | That column does not exist (`42703`). An account's zkApp is reached through `accounts_accessed.zkapp_id`. |
| `SELECT DISTINCT … ORDER BY random()` | Postgres rejects it (`42P10`). |

The year assertions are replaced by `assertPlausibleBlockTime`, which checks the timestamp parses, falls after 2020, and is not in the future — an invariant that survives an hourly-rotating dump, which is what the rest of this suite already asserts.

## Testing

Run end to end against the real **2026-09-02 20:00 devnet dump** (1.4 GB, 692,089 blocks, canonical heights 1 → 552,691), loaded into `postgres:17` to match CI:

```
# Dump loaded: 692089 blocks in archive.
# tests 25
# suites 10
# pass 25
# fail 0
```

All ten suites pass, including the ones that had never executed a single assertion: `Quickcheck: random zkApp addresses`, `Events and Actions`, and `Action-state invariants on real data`.

The corrected SQL was checked directly rather than reasoned about — the `accounts_accessed.zkapp_id` join returns real zkApp addresses from the dump, and both query shapes were run against a minimal schema replica to confirm the old one errors and the new one returns rows.

`npm run build` and `npm run lint` are clean. Prettier reports this file as unformatted both before and after these changes; the hunks it wants are in pre-existing code, so I left the formatting alone.

## Note

This does not make the nightly *newly* green — it makes it run for the first time. Anything it reports from here is a real finding about real data.

https://claude.ai/code/session_012LtdRKTkaKRAuTkEEpgPcc


tests/devnet-dump/setup.ts
```diff
@@ -48,37 +103,60 @@ export async function downloadDump(): Promise<string> {
...
-  } finally {
-    await admin.end();
+  // A `--create` dump brings its own database and ignores psql's `-d`; a plain
+  // dump needs one made for it. Load into `postgres` in the first case so the
+  // dump's own CREATE DATABASE runs against a connection it is not replacing.
+  const dumpDb = databaseCreatedByDump(dumpPath);
+  targetDb = dumpDb ?? PG_TEST_DB;
+  const loadDb = dumpDb ? 'postgres' : PG_TEST_DB;
+
+  await dropDatabase(targetDb);
```
SanabriaRusso, line 113: This is the footgun: `targetDb` is now `archive` for every real devnet dump, and `archive` is also the conventional database of a local Mina archive node. Someone running this suite against their workstation Postgres to reproduce a nightly failure drops their archive without a prompt, and again in `teardownTestDatabase`.

Two ways out, either is fine by me:

1. **Keep the harness in charge of the name.** The `--create` prelude is three lines at the top of the file. Filter them out and load into `PG_TEST_DB` as before, e.g. `sed` restricted to the first ~80 lines deleting `^CREATE DATABASE `, `^ALTER DATABASE ` and `^\connect `, piped into `psql -v ON_ERROR_STOP=1 -d ${PG_TEST_DB}`. Limiting the edit to the head keeps it away from COPY data. This removes `databaseCreatedByDump`, the mutable `targetDb`, and the "never interpolate a downloaded name into DROP DATABASE" concern entirely, and `PG_TEST_DB` keeps meaning what the README says.

2. **Keep this approach, but make the drop opt-in when the name came from the dump.** For example refuse unless `PG_TEST_DB` was explicitly set to the same name, or unless a `DEVNET_DUMP_ALLOW_DROP` flag is present, and document it in the README's configuration table.

If you prefer to land (2) later, a README line saying "this drops and recreates a database named `archive` on `PG_TEST_HOST`" is the minimum for this PR.

tests/devnet-dump/setup.ts
```diff
@@ -48,37 +103,60 @@ export async function downloadDump(): Promise<string> {
...
+      `PGPASSWORD=${PG_TEST_PASSWORD} psql -v ON_ERROR_STOP=1 -h ${PG_TEST_HOST} -p ${PG_TEST_PORT} -U ${PG_TEST_USER} -d ${loadDb} -f ${dumpPath}`,
+      { stdio: 'pipe', timeout: 600000 }
     );
-    await admin.unsafe(`DROP DATABASE IF EXISTS ${PG_TEST_DB}`);
+  } catch (error) {
+    // Without this the loader swallowed psql's stderr and reported success,
+    // which turned a failed load into a confusing `relation "blocks" does not
+    // exist` in every test.
+    const stderr = (error as { stderr?: Buffer }).stderr?.toString() ?? '';
+    throw new Error(`Loading the devnet dump failed:\n${stderr.trim()}`);
```
SanabriaRusso, line 137: When `execSync` fails for a reason other than a psql error, `stderr` is usually empty: the 600 s `timeout`, or `maxBuffer` (default 1 MiB, and `stdio: 'pipe'` buffers psql's stdout command tags). The thrown message is then just `Loading the devnet dump failed:` with nothing after it. Appending `(error as Error).message` covers both, and `-q` on the psql line keeps stdout from ever reaching the buffer limit as the dump grows.

tests/devnet-dump/devnet-dump.test.ts
```diff
@@ -59,6 +59,31 @@ after(async () => {
...
 
+/**
+ * A block timestamp is plausible when it is not epoch 0 and not in the future.
+ *
+ * The suite asserts invariants rather than values, because the dump rotates
+ * hourly. An earlier version pinned the year at >= 2023, which no dump can
+ * satisfy: devnet's genesis block is dated 2021-09-24, so the oldest blocks in
+ * every dump fail it.
+ */
+const MINA_EPOCH_START = Date.UTC(2020, 0, 1);
```
SanabriaRusso, line 70: Nit: "epoch" already means something specific in Mina consensus, and 2020-01-01 predates both mainnet (2021-03-17) and devnet (2021-09-24) genesis. `EARLIEST_PLAUSIBLE_BLOCK_TIME` would say what the constant is for.

SanabriaRusso (approved):
**Approve.** The diagnosis is correct, the fix is verified, and this is the piece that lets the nightly finally say something about real post-Mesa data.

### What I verified
- **The dump really is `pg_dump --create`.** I range-fetched the head of `devnet-archive-dump-2026-09-04_0500.sql.tar.gz` from GCS: `-- Dumped by pg_dump version 17.4`, `CREATE DATABASE archive WITH TEMPLATE = template0 ... LOCALE = 'en_US.UTF-8'` at line 24, `\connect archive` at byte 652, and no `OWNER TO` / `ALTER DATABASE` statements. So the 64 KB header read is more than enough, the plain-identifier regex matches the real name, and `ON_ERROR_STOP=1` will not trip on ownership statements against the CI `postgres:17` service.
- **Schema claims.** `tests/integration/fixtures/archive_db.sql` agrees: `zkapp_accounts` has no `account_identifier_id`; `accounts_accessed.zkapp_id` exists and is nullable. The `SELECT DISTINCT ... ORDER BY random()` rejection is a real Postgres rule (`42P10`).
- **CI on this branch.** The `workflow_dispatch` run of *Nightly Devnet Dump Tests* on `fix/devnet-dump-suite` (run 33691796223, 2026-09-02) loaded 692,116 blocks into `archive` in ~78 s and finished **25/25 pass across 10 suites**, including the three suites that had never executed an assertion before. Every scheduled run on `main` before and since is red with the old `relation "blocks" does not exist`.
- **Relation to #229 (Mesa best-chain anchor).** No overlap: this PR touches only `tests/devnet-dump/`, #229 touched `src/` and `tests/integration/`. The branch is based just before #229, so the green run above exercised the *old* events/actions anchor. The devnet dump as of 2026-09-02 is two weeks past devnet's fork and the new chain has long outgrown the abandoned one, so the anchor change is a no-op on that data. Still, I'd trigger the nightly by hand right after merging rather than waiting for 03:30 UTC, so the first run on `main` carries both changes.
- `tsconfig` includes `tests/**`, so the `Run-Tests` check compiling this file confirms the `connectionString` -> `getConnectionString()` rename has no other importers (the other `connectionString` exports live in `tests/integration/*` and are separate constants).

### One thing I'd like addressed (follow-up is fine)
`setupTestDatabase` now runs `DROP DATABASE archive` on whatever server `PG_TEST_HOST` points at, and `archive` is the default database name of a locally running Mina archive node. A developer who runs `npm run test:devnet-dump` against their local Postgres to reproduce a nightly failure silently loses their archive. Before this PR the harness only ever dropped `devnet_dump_test`. See the inline comment for the two options I see; at minimum the README and the header docstring should say which database gets dropped.

### Nits (inline)
- The catch around `psql` loses the reason when the failure is not a psql error (timeout, `maxBuffer`).
- `MINA_EPOCH_START` reads as a consensus-epoch constant; something like `EARLIEST_PLAUSIBLE_BLOCK_TIME` says what it is.

### Suggested follow-up, not for this PR
The suite never queries with `inBestChain: true`, so it would not have caught the #229 hard-fork shape even on a dump taken during that window. A dump-agnostic invariant would: `blocks({blockHeight_gt: canonicalMax, inBestChain: true})` must reach `networkState.pendingMaxBlockHeight`, and every block above the pending tip must be `orphaned`. That is the check that was missing on 2026-09-03.

PR #233, opened 2026-09-14
ci: publish the container for linux/arm64 as well as linux/amd64

dkijania:
## Why

`ghcr.io/o1-labs/archive-node-api` currently ships **one architecture**. Its manifest index lists only:

```
linux amd64
unknown unknown   <- build provenance attestation, not an architecture
```

`docker/build-push-action` was called with no `platforms:`, so it built for the `ubuntu-latest` runner only. On Apple Silicon that means emulation and a platform-mismatch warning, or a flat `no matching manifest for linux/arm64` when emulation is off.

## What changed

- **Set up QEMU** before Buildx, so the amd64 runner can execute aarch64 build steps.
- `platforms: linux/amd64,linux/arm64` on the build step.

No `Dockerfile` change is needed: I checked the pinned base digest against Docker Hub and `node:22-alpine@sha256:16e22a55…` is a multi-arch **index** that already carries `linux/arm64/v8`, alongside amd64, arm/v6, arm/v7 and s390x. The digest pin and its reproducibility guarantee stay exactly as they were.

## The cost, stated plainly

arm64 layers are built under emulation on an amd64 runner, so this job gets slower — `npm ci` and `tsc` are the expensive parts. If that becomes painful, the alternative is a matrix over `ubuntu-latest` and `ubuntu-24.04-arm` that builds each architecture natively and then merges the manifests. This repository is public, so the arm runners cost nothing. I kept this PR to the smaller change; say the word and I will do the matrix instead.

## How to verify after merge

```sh
docker buildx imagetools inspect ghcr.io/o1-labs/archive-node-api:latest
```

should list both `linux/amd64` and `linux/arm64`.

https://claude.ai/code/session_01KfPTtp9B879CdXFu2hSU7d


dkijania:
CI is green, and the emulation cost is now measured rather than guessed.

`build-and-deploy` on this branch took **18m41s**. The same job on #232, which is also based on current `main` but builds amd64 only, took **6m04s**. So adding emulated arm64 costs about **+12m30s**, roughly tripling that job. Every other check is unchanged.

The arm64 layers do build and push, so the approach works — this is purely a wall-clock question.

If 18 minutes is too slow for the main branch, the alternative I mentioned in the description is a matrix over `ubuntu-latest` and `ubuntu-24.04-arm` that builds each architecture natively and merges the manifests. It is a larger diff, but it would bring the job back to roughly single-architecture time. Happy to switch this PR to that shape if you prefer it.

Note that PR builds push dev tags to GCP Artifact Registry only, so the multi-arch manifest cannot be inspected on ghcr.io until this lands and a release tag is built.

SanabriaRusso:
## Review

The PR does what it says, and every factual claim in the description checks out. I verified rather than read:

- **The pin is correct.** `gh api repos/docker/setup-qemu-action/git/ref/tags/v4.3.0` → `1f40c72289eff860ee54a304f1438e3cff362e0a`. Exact match. (I also re-checked the neighbouring `setup-buildx-action` pin `37fe631…` against its `v4.3.0` tag — also exact. The file's pin comments are truthful.)
- **`platforms: arm64` is the documented form**, not a typo for `linux/arm64`. `action.yml` at that SHA documents the input as "Platforms to install (e.g. `arm64,riscv64,arm`)". The run log shows binfmt registering `linux/arm64` correctly.
- **The base-image claim is true.** Queried the Docker Hub registry API directly: `node:22-alpine@sha256:16e22a55…` is an OCI image *index* carrying `linux/arm64/v8` (`sha256:d51cff3f…`) alongside amd64, arm/v6, arm/v7 and s390x. Both Dockerfile stages use that identical digest, so there is no second base image to catch us out. The digest pin and its reproducibility guarantee are unaffected.
- **Nothing downstream breaks.** `build.yaml` has exactly one job and nothing after the build reads `steps.build.outputs.digest` or `imageid`. The SBOM job in `security.yaml` scans the source tree (`path: .`), never the image. `smoke-load` lives in a separate workflow, builds from source, and never pulls the image — it was absent from this PR's checks because of path filters, not because of anything here.

So: **no blocking defects.** What I would like to resolve before merge is the cost, because measurement shows it is being paid on the wrong events and for the wrong work.

---

### 1. Every PR pays the 3.3x, but only tag builds reach GHCR

The trigger is `pull_request: branches: [main]` plus `push: tags: ['v*']`, with no `if:` gating the new platform list — so both architectures are built on every same-repo PR.

Meanwhile `Determine tags` (`.github/workflows/build.yaml:209-243`) emits GHCR tags **only** on a tag push. PR builds push `dev-<sha>` to GAR only:

```
# PR build - use dev tags with SHA (GCP only)
```

The 13 extra minutes on every PR therefore produce an arm64 layer that never reaches `ghcr.io/o1-labs/archive-node-api` — the registry this PR exists to fix.

There is a genuine counter-argument, and I want to state it rather than skip it: building arm64 on PRs is what proves the arm64 build still works *before* you cut a tag. Gating it to tags means a broken arm64 build is discovered at release time, which is the worst moment. I do not think the answer is to give up that signal — I think it is to make the signal cheap, which is finding 2.

### 2. The real cost is `npm ci` under emulation — and most of it should not be running at all

Per-step timings from run `34880710868`:

| Dockerfile step | amd64 | arm64 (QEMU) | ratio |
|---|---|---|---|
| `RUN npm ci` | 111.8s | **752.2s** | **6.7x** |
| `RUN npm run build` (tsc) | 4.6s | 40.6s | 8.8x |
| `RUN apk add --no-cache tini` | 0.9s | 3.6s | 4.0x |

The emulated `npm ci` alone is **~89% of the 847s build graph** and sits on the critical path.

`Dockerfile:7` is a bare `npm ci`, so the image build installs `artillery`, `o1js`, `typescript`, `eslint` and `@playwright/browser-chromium` — and runs their install scripts — then `Dockerfile:18` `COPY --from=build /app/node_modules` ships every one of them into the **runtime image**. Under emulation that means a browser download and a `unix-dgram` node-gyp compile executed through qemu-aarch64. They didn't break (exit 0), but they are a latent breakage source and nearly all of those 752 seconds.

Two facts make this cheap to fix:

- **tsc output is architecture-independent**, so the compile stage can be pinned to the build platform and run natively once for all targets.
- **No production code imports a devDependency.** I checked: `grep -rn "from '(o1js|graphql-request|ts-node|cross-env|artillery|c8)" src/` returns nothing, and the entrypoint is `node build/src/index.js`. `--omit=dev` is safe for the runtime image.

Suggested `Dockerfile`:

```dockerfile
# Stage 1: runtime dependencies — TARGET platform, production only.
FROM node:22-alpine@sha256:16e22a550f3863206a3f701448c45f7912c6896a62de43add43bb9c86130c3e2 AS deps
WORKDIR /app
COPY package*.json ./
RUN npm ci --omit=dev --ignore-scripts

# Stage 2: compile. tsc output is architecture-independent, so pin this stage
# to the BUILD platform — it runs natively once and serves every target.
FROM --platform=$BUILDPLATFORM node:22-alpine@sha256:16e22a550f3863206a3f701448c45f7912c6896a62de43add43bb9c86130c3e2 AS build
WORKDIR /app
COPY package*.json ./
RUN npm ci --ignore-scripts
COPY src ./src
COPY tsconfig.json ./
RUN npm run build

# Stage 3: runtime — unchanged except these two COPYs.
COPY --from=deps  /app/node_modules ./node_modules
COPY --from=build /app/build ./build
```

**The trap, flagged explicitly:** with `--platform=$BUILDPLATFORM` on the build stage you must **not** keep `COPY --from=build /app/node_modules` — that would put amd64-resolved `node_modules` into the arm64 image. The separate `deps` stage is what makes this safe. Please keep the two `COPY --from=` lines exactly as written above.

This also shrinks the runtime image and drops the playwright browser and the node-gyp compile from it entirely — worth having on its own merits, independent of arm64.

**Acceptance criteria**
- `build-and-deploy` on a PR returns to under ~9 minutes with both platforms still built.
- `docker buildx imagetools inspect <image>` lists `linux/amd64` and `linux/arm64`.
- `docker run --platform linux/arm64 <image>` serves `GET /healthcheck` with 200.
- The runtime image no longer contains `node_modules/typescript` or `node_modules/@playwright`.

**How to test**
```sh
docker buildx build --platform linux/amd64,linux/arm64 -t ana:multiarch .
docker run --rm --platform linux/arm64 -e PG_CONN=... -p 8080:8080 ana:multiarch &
curl -fsS localhost:8080/healthcheck
docker run --rm --platform linux/arm64 ana:multiarch ls node_modules/typescript   # expect: not found
```

If you would rather keep this PR strictly to the workflow, the fallback is to gate the platform list to tag builds — but I'd take the Dockerfile fix, since it keeps the PR-time arm64 signal that finding 1 would otherwise cost us.

### 3. On the matrix alternative you offered

I'd skip it. Native arm runners are free on a public repo, but splitting this job means duplicating WIF auth and GAR login across both legs — which widens credential exposure against the grain of the `ORDERING RULE` comment at `:35-36` and issue 03 — or restructuring so `npm publish` stays in one place. Realistically 100+ lines with regression risk on exactly the credential-hygiene invariants this job was hardened for. Your instinct that the matrix is the escape hatch is right in spirit, but it aims at the wrong bottleneck: the fix is to stop doing the expensive work, not to find a faster machine to do it on.

---

### Separately: the buildx cache has never hit once (pre-existing, not yours)

Worth an issue rather than a change here. From this PR's log:

```
Cache not found for input keys: Linux-buildx-deb9e9b3…, Linux-buildx-
```

Every buildx cache entry in the repo is scoped to `refs/pull/N/merge` — there is not one on `refs/heads/main`. `build.yaml` has no `push: branches: [main]` trigger, so the base-branch scope is never written, and GitHub scopes a PR-created cache to that PR's own ref, invisible to every other PR. The repo has been writing ~691 MB per PR into a cache only a *second run of the same PR* could read.

This PR doubles that to **1.39 GB** (measured via `gh api .../actions/caches`: 691,198,082 B on #234 vs 1,390,076,079 B here — exactly 2.01x), plus a new 32 MB binfmt entry. The repo sits at **4.23 GB of the 10 GB limit**; at 1.39 GB per PR, ~7 concurrent PRs start LRU-evicting — and the victims include the `node-cache-…-npm` entry on `refs/heads/main` that lint, unit-tests and smoke-load depend on.

Two things, both follow-ups:
- Make the key platform-aware so single- and multi-arch entries can't cross-contaminate.
- Either add `push: branches: [main]` so the base-branch scope gets populated and the cache starts earning its keep, or delete the cache steps — right now they are pure cost.

Also a nit while you're in there: `Move cache` lacks `if: always()`, so on a failed build the rotation is skipped and the post step saves the stale cache under the new key.

---

**Disposition:** approving the approach — the change is correct and the pins are clean. I'd like finding 2 in before merge (it's ~10 lines of `Dockerfile` and it's what makes arm64 affordable). Finding 3's cache work belongs in its own issue.

dkijania:
Finding 2 is in, as `17885ca`, and the branch is rebased onto current `main`. Finding 3 is filed separately (link below). Everything here is measured, not estimated.

## The Dockerfile now has three stages

Exactly the shape you specified, including the trap you flagged — `node_modules` comes from `deps`, never from the `$BUILDPLATFORM`-pinned `build` stage.

One thing I checked before taking `--ignore-scripts`, since it is the part that could bite silently: **no production dependency has an install script.** The node-gyp compile you saw is `unix-dgram`, reached only through `artillery` → `artillery-plugin-publish-metrics` → `hot-shots`, and the lockfile marks it `optional: true, dev: true`. `--omit=dev` removes it outright, so there is nothing for the ignored scripts to have built.

## Measured, amd64, cold cache

| | before | after |
|---|---|---|
| image size | 1,297,918,614 B (1.30 GB) | **221,929,240 B (222 MB)** |
| runtime `node_modules` | 780.7 MB | **54.7 MB** |

And the number your acceptance criterion asks about — a **cold build of both architectures took 2m30s**, against **5m28s** for the old *single*-architecture build on the same machine. Dual-arch is now faster than single-arch was. CI will have its own numbers, but the direction is not in doubt.

## Acceptance criteria, run

- ✅ Both platforms in one manifest: a `--platform linux/amd64,linux/arm64` build emits an OCI index listing `linux/amd64` and `linux/arm64`.
- ✅ arm64 image serves: started under emulation, `/healthcheck` **200**, `/readiness` **200**, `POST / {__typename}` → `{"data":{"__typename":"Query"}}`, tini still PID 1, `process.arch` = `arm64`, zero native `*.node` files.
- ✅ Dev tree gone from the runtime image: `typescript`, `artillery`, `eslint` and `o1js` all absent.

**One correction to that last criterion.** `node_modules/@playwright` still *exists* — as an empty scope directory, 4.0K, nothing inside. npm leaves the scope behind when it omits the package. The package is gone; the directory is not. Saying the criterion passed without that caveat would have been inaccurate.

## Also

Added `if: always()` to `Move cache`, your nit, plus a guard so the `mv` cannot fail the step when the build produced no new cache directory.

## Finding 3

Filed as its own issue rather than changed here, as you suggested — see the link in the next comment.

dkijania:
Finding 3 is now #241 — the buildx cache is written on every PR and read by nothing, and this PR doubles what it writes. Options laid out there; nothing needed in this PR for it.

dkijania:
CI numbers are in, and they beat the acceptance criterion by more than expected.

`build-and-deploy` on this branch: **4m37s**.

For comparison, all on this same job:

| build | time |
|---|---|
| this branch **before** the Dockerfile fix (amd64 + arm64) | 18m41s |
| `main`, amd64 only (measured on #232) | 6m04s |
| **this branch now (amd64 + arm64)** | **4m37s** |

So building both architectures is now faster than building one was — the emulated `npm ci` was costing more than the entire rest of the job. Your acceptance criterion was under ~9 minutes with both platforms; it lands at roughly half that.

All seven checks pass: Check Schema 18s, Linting 1m29s, Run-Tests 14m28s, SBOM 9s, build-and-deploy 4m37s, npm audit 1m51s, unit-tests 1m59s.

dkijania:
Rebased onto `main` after #239 merged, and the base moves with it.

#239 bumped the image to `node:25-alpine`, which is **end-of-life** — v25 is an odd-numbered line, never LTS, `end: 2026-06-01`. So this branch now pins all three stages to `node:24-alpine@sha256:ebfe2f90…`, Active LTS until 2028-04-30, matching #235. That way `main` stops shipping an unsupported runtime whichever of the two lands first.

Re-verified after the rebase, on Node 24:

- cold `--platform linux/amd64,linux/arm64` build: **3m13s**, both platforms in the manifest;
- arm64 image runs `v24.21.0`, `process.arch` = `arm64`;
- runtime `node_modules` 54.5 MB, with `typescript`, `artillery`, `eslint` and `o1js` all absent.

**Overlap with #235, stated plainly:** both PRs now set the same `FROM` digest, so whichever merges second will show a small conflict on those lines. It resolves by keeping this PR's three-stage structure with that same Node 24 digest — no judgement call in it. Everything else is disjoint: #235 owns the CI matrix, the EOL Node 20 workflow fixes and the dependabot rule; this PR owns QEMU, `platforms:` and the Dockerfile stages.

PR #231, opened 2026-09-14
docs: badge the container image and the three client SDKs

dkijania:
The README already advertised the npm package. This adds:

- a badge for the GHCR container image (`ghcr.io/o1-labs/archive-node-api`),
- a **Client SDKs** row with version badges for the TypeScript, Rust and Go SDKs, each linking to its registry.

A reader can now reach every published artifact of the 1.x release from the hub repository.

All badge image URLs were checked and return SVG.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01KfPTtp9B879CdXFu2hSU7d

PR #230, opened 2026-09-09
ci: authenticate to Google Cloud keylessly with WIF

SanabriaRusso:
## Summary

`GCP_SA_KEY` — the long-lived Google service-account key this workflow used — leaked and has been disabled, so `Docker Build CI/CD` currently fails at the auth step. This replaces it with **Workload Identity Federation** and rewrites the job so a live credential can never reach dependency code again.

- **Keyless auth.** `google-github-actions/auth` now takes `workload_identity_provider` + `service_account` instead of `credentials_json`, with `token_format: access_token`, `create_credentials_file: false` and `access_token_lifetime: 900s`. There is no `GCP_SA_KEY` anywhere in `.github/` any more, and no credential file is written into `$GITHUB_WORKSPACE`.
- **Scoped, not global.** The workflow-level `env.GCP_SA_KEY` (which exported the raw key into *every* step, including `npm ci`'s install scripts) is gone. The minted token is referenced only as `steps.auth.outputs.access_token`, by the GAR login, the ghcr login and `npm publish`.
- **Reordered.** `npm ci --ignore-scripts` and `npm run build` now run **before** authentication. After the auth step only the registry logins, `npm publish --ignore-scripts` and the docker build remain. An `ORDERING RULE` comment at the top of the job records why.
- **No credentials on disk.** `gcloud auth configure-docker` is replaced by `docker/login-action` (writes `~/.docker/config.json`, logs out in its post step); the hand-rolled `echo … > .npmrc` / `rm -f .npmrc` pair is replaced by `actions/setup-node` with `registry-url` + `scope: '@o1-labs'`, which keeps its `.npmrc` in `$RUNNER_TEMP`, plus a per-step `NODE_AUTH_TOKEN` on the publish step only.
- **Guard step**, run twice — right before `npm publish` and again right before the docker build context is sent — failing the job if any `gha-creds-*.json`, `.npmrc` or dotenv file is in the workspace root, or if `npm pack` would ship one.
- **Ignore files.** `gha-creds-*.json` added to `.gitignore`, `.dockerignore` and `.npmignore`; `.npmrc` added to `.gitignore`.
- **Hardening (issue 06, workflow part only).** Every workflow now declares a top-level `permissions:` block, and every non-local `uses:` is pinned to a 40-char commit SHA with the tag kept as a trailing comment so Dependabot's `github-actions` group still works.

Tag formats, image names, registries, the npm version scheme, the fork/Dependabot `if:` guard and the buildx layer cache are all unchanged.

[...]

Issue #208, opened 2026-07-30
npm publishing from CI has never worked: trusted publisher is not configured (v0.0.9 tagged but unpublished)


SanabriaRusso:
## Current state

`v0.0.9` is tagged at `e353bbc` and **is not published**. npm still serves only `0.0.6`. The tag is valid and the workflow file at that commit is correct — the only missing piece is on the npm side.

Not urgent: the Docker image is the artifact currently being consumed, so this blocks the npm channel only.

## Symptom

[Run 30543971082](https://github.com/o1-labs/Archive-Node-API/actions/runs/30543971082) got all the way through `Update npm`, `npm ci`, `npm run build`, and a full green `npm test`, then failed on the last step:

```
npm error code ENEEDAUTH
npm error need auth This command requires you to be logged in to https://registry.npmjs.org/
```

The tarball itself built fine — `0.0.9`, 136 files, 100.2 kB packed / 524.2 kB unpacked.

## Root cause

`publish-npm.yml` depends entirely on OIDC trusted publishing. It declares `permissions: id-token: write`, but:

- it sets no `NODE_AUTH_TOKEN`, and `actions/setup-node@v4` is configured without `registry-url`, so no `.npmrc` is ever written — there is no token fallback;
- the repository has no npm credential of any kind (`gh secret list` returns only `GCP_SA_KEY`).

So the publish can only succeed if npm accepts the OIDC token via a **trusted publisher configured on the package**. That configuration does not exist, hence `ENEEDAUTH`.

## This has never worked

Worth recording explicitly, because the workflow's presence implies otherwise:

- `0.0.6` was published at `2026-04-27T22:48:59.899Z` by npm user `o1-labs <npm@o1labs.org>`, and the package's `created` timestamp is the same instant.
- No CI run exists at that time. The nearest successful `Publish Package` run was `20344320453` on 2025-12-18, four months earlier, from a `workflow_dispatch` on branch `leon/fix-release-workflow`.
- Every tag-triggered publish run in this repo's history has failed: `v0.0.6` ([25001081251](https://github.com/o1-labs/Archive-Node-API/actions/runs/25001081251)), `v0.0.9` first attempt ([30540591005](https://github.com/o1-labs/Archive-Node-API/actions/runs/30540591005), a separate Node-engine bug fixed in #207), and `v0.0.9` second attempt ([30543971082](https://github.com/o1-labs/Archive-Node-API/actions/runs/30543971082), this issue).

[...]

PR #234, opened 2026-09-15
fix(config): let ENABLED_QUERIES name verificationKeyUpdates, release 1.0.1

dkijania:
## What

`KNOWN_QUERIES` in `src/config.ts` is the hand-maintained allowlist that `validateConfig` checks `ENABLED_QUERIES` against. It still listed only `events, actions, networkState, blocks`. `verificationKeyUpdates` joined `schema.graphql` and was never added.

Effect: an operator who sets `ENABLED_QUERIES=verificationKeyUpdates` gets

```
ENABLED_QUERIES contains unknown queries: verificationKeyUpdates.
```

and the process exits at startup, for a field the server really serves. Operators who leave `ENABLED_QUERIES` unset are unaffected.

The runtime filter in `resolvers.ts` derives its field list from the schema AST, so only the validator was wrong — nothing was actually unserveable.

## Changes

- `src/config.ts` — add `verificationKeyUpdates` to `KNOWN_QUERIES`.
- `tests/unit/config.test.ts` — new test that parses `schema.graphql`, reads the root fields of `type Query`, and asserts `validateConfig` accepts each one. The next root field added to the schema now fails the suite instead of shipping. Confirmed it fails without the one-word fix.
- `docs/getting-started.md` — the `ENABLED_QUERIES` row listed the same stale four names.
- `package.json` / `package-lock.json` — bump to `1.0.1`. The lockfile still recorded `"license": "ISC"`; #232 changed only `package.json`.

## Release

This is the commit to tag `v1.0.1` and publish. npm still serves `0.0.6`; the Trusted Publisher for `@o1-labs/mina-archive-node-graphql` is now registered, so `publish-npm.yml` can authenticate over OIDC. Tagging `v1.0.0` again was the alternative, but that tag predates #230, #231 and #232, so its tarball would ship without the Apache-2.0 `LICENSE` and would declare ISC.

`npm pack --dry-run` on this branch: 217 files, `LICENSE` included, version `1.0.1`.

## Test

- `npm run test:unit` — 194/194 pass.
- `npm run lint` — clean.


SanabriaRusso (approved):
## Review

**The fix is correct, minimal and provably tested. Approving.** The notes below are about the release half, which rides along in the same PR and has one precondition I think is still unmet — but none of it is a reason to hold the bug fix.

### The code change verifies out

- **The "five root fields" claim is exact.** Parsed `schema.graphql` with the repo's own `graphql` package: `events, actions, verificationKeyUpdates, networkState, blocks`. `KNOWN_QUERIES` now lists those five **in the same order**, with nothing missing and nothing extra. `verificationKeyUpdates` arrived in #225 and the allowlist was never updated.
- **The test genuinely fails without the one-word fix.** Reverted only `src/config.ts` to main's version and ran `npm run test:unit`:
  ```
  ✖ accepts every root query field declared in schema.graphql
  AssertionError: ENABLED_QUERIES=verificationKeyUpdates should be accepted; add it to KNOWN_QUERIES
  ℹ fail 1
  ```
  The assertion message names the offending field *and* the remedy. That's the right way to write it.
- **The PR body's "nothing was actually unserveable" claim is right.** `src/resolvers.ts:88-101` filters from the schema AST and `:36-44` already registers the `verificationKeyUpdates` resolver, so the defect was purely in the startup validator. End-to-end with `ENABLED_QUERIES=verificationKeyUpdates,blocks` the server exposes exactly those two fields.
- **Deriving from the schema is the right shape of fix.** The alternative — `as const satisfies readonly (keyof Query)[]` off `src/resolvers-types.ts` — is free at runtime but only guards if someone remembered to run `npm run codegen`. Reading `schema.graphql` catches the forgot-to-run-codegen case too, and `unit-tests.yaml` has no path filters so it fires on schema-only PRs. Good call.
- **Docs are complete and correctly scoped.** `docs/getting-started.md:202` now matches the schema order. `README.md:69` says "subset of root query fields" without enumerating, so it correctly needed no change — exactly the AGENTS.md doc rule.
- **The lockfile licence change is complete.** It touches only the root `packages[""]` entry. The remaining hundreds of `"license": "ISC"` hits are third-party dependency licences and must *not* be touched.
- `npm run test:unit` 194/194 and `npm run lint` clean, reproduced locally.

---

### Release: one correction, one precondition, one contradiction

**1. I think the premise in "Release" is the wrong blocker — and the good news is your own analysis already found the right one.**

The PR says the Trusted Publisher is now registered "so `publish-npm.yml` can authenticate over OIDC". But your comment on #208 (2026-09-05) established that the v1.0.0 run failed with `ENEEDAUTH`, which means npm was offered **no credential at all** — the request never got far enough for the npm-side registration to matter. Registering the Trusted Publisher doesn't by itself clear an `ENEEDAUTH`.

**That said, I believe #208's conclusion is now stale, in your favour.** #208 blames the org Actions policy silently dropping `id-token: write`. But #230 landed WIF on 2026-09-09 — five days *after* that v1.0.0 run — and `build.yaml` now mints an OIDC token successfully on every PR. On this PR's own run (`34965907056`), `google-github-actions/auth` ran with the WIF provider and the job went on to push to GAR, which is impossible without a minted OIDC token. The Actions token policy is per-repository, not per-workflow, so whatever was missing on 2026-09-04 is evidently not missing now.

So the honest state is: **both halves are now plausible and neither has been exercised.** The failure modes are cheaply distinguishable, which is what makes this safe to just try:

| Symptom on the v1.0.1 tag | Meaning |
| --- | --- |
| `ENEEDAUTH` | still no OIDC token — GitHub-side, reopen #208 |
| `403` from the registry | token sent and rejected — npm-side registration has the wrong repo/workflow filename |
| success | done |

Worth updating #208's title and diagnosis either way; it currently sends the next person to npmjs.com.

**2. Please don't "fix" this by re-publishing at the existing `v1.0.0` tag.** Recording this because it's the obvious-looking recovery and it's wrong: `v1.0.0` points at `18b2ef9`, which predates #230/#231/#232, so its tarball ships **without the Apache-2.0 `LICENSE` and declaring ISC**. Your reasoning for 1.0.1 over re-tagging is sound. Note npm would then go `0.0.6 → 1.0.1` with 1.0.0 never existing on the registry — that's fine (1.0.0 is unclaimed, and GHCR already carries `1.0.0`/`1.0`/`1`/`latest` from that tag's successful image build), but `docs/versioning.md`'s "Migrating from npm 0.0.6" section names `1.0.0` and should be reworded to 1.0.1.

**3. This PR contradicts the documented release flow — worth one line in this PR.** `docs/versioning.md:123-132` tells maintainers to run `npm version <patch>`, which bumps `package.json` *and* creates the tag. The carve-out for bumping in advance is explicitly scoped to "the initial `1.0.0` release only". Follow the documented flow after merging this and you land on **1.0.2**, silently skipping 1.0.1.

Bumping in the PR *is* the right mechanism — I verified neither pipeline derives the version from the tag (`publish-npm.yml` runs plain `npm publish`, and `build.yaml:74-78` says `# Release build - use version from package.json`), so `package.json` must carry the version before the tag exists. The docs just need to say so.

**Suggested change** to `docs/versioning.md:130-132`:
```diff
-For the initial `1.0.0` release only, `package.json` on `main` already carries
-the version to release. Tag it directly (`git tag v1.0.0 && git push
---follow-tags`) rather than running `npm version`, which would bump past it.
+When `package.json` on `main` already carries the version to release — as it
+does when the bump landed in an earlier PR — tag it directly
+(`git tag vX.Y.Z && git push --follow-tags`) rather than running `npm version`,
+which would bump past it.
```

**Acceptance criteria**
- `docs/versioning.md` no longer restricts the tag-directly path to 1.0.0, and the migration section names the version actually published.
- After merge, `git tag v1.0.1 && git push --follow-tags` publishes `1.0.1` — not 1.0.2.
- `npm view @o1-labs/mina-archive-node-graphql versions` lists `1.0.1`.

**How to test the release before committing to it:** push the tag and watch `publish-npm.yml`. A failure is fully recoverable — `1.0.1` stays unclaimed and the job re-runs from the same tag once the cause is cleared. Confirm `Id-Token: write` appears in the run's `GITHUB_TOKEN Permissions` group before assuming the npm side is at fault.

---

### Nits, none blocking

- **`tests/unit/config.test.ts:18`** — `path.resolve(process.cwd(), 'schema.graphql')` is cwd-coupled. Fine for every documented invocation (npm sets cwd to the package root, and CI runs `npm run test:coverage` from the root), and it fails loudly with `ENOENT` rather than passing falsely. But it breaks from any other cwd. Mirroring `src/resolvers.ts:12-14` would fix it:
  ```ts
  const here = path.dirname(fileURLToPath(import.meta.url)); // build/tests/unit
  const schemaPath = path.resolve(here, '../../../schema.graphql');
  ```
  *Acceptance:* `cd build && node --test tests/unit/config.test.js` passes.
- **`tests/unit/config.test.ts:12-16`** — the comment says it resolves "against the source tree rather than the test's own location", but it resolves against cwd. Reword if you touch it.
- **Optional hardening** for `src/config.ts:22-28` — adding `as const satisfies readonly Exclude<keyof Query, '__typename'>[]` gives the *reverse* check (a typo'd or stale name fails `npm run build`) for one line. The runtime test already covers the forward direction.
- **Duplicates in `ENABLED_QUERIES` are accepted silently.** Verified harmless: `"blocks, blocks ,blocks"` exposes `blocks` once. Not worth code.
- **Pre-existing, cosmetic:** `src/resolvers.ts:73`'s `if (enabledQueries.length > 0)` is dead — `''.split(',')` is `['']`, length 1, so it can never be false.
- **Pre-existing, worth its own issue:** nothing checks the pushed tag against `package.json#version`. Tag `v1.0.2` with `package.json` at `1.0.1` and you silently publish 1.0.1 under the wrong tag. A four-line guard in `publish-npm.yml` before the publish step closes it.

---

### On the SDK audit

For the record, this PR closes **Part 5 of the SDK audit brief** — the one finding that brief explicitly assigns to this repo rather than to any of the three client SDKs. The schema-derived test means the same drift can't recur. I checked the rest of the brief against current main; the other server-side items it implies are unaddressed by either open PR and none are tracked yet. I'll file those separately — the highest-leverage one by far is tightening the eight `[T]!` list positions to `[T!]!`, which is verified non-breaking by the repo's own pinned `graphql-inspector` gate and would retire the three SDK issues that otherwise force three client major releases.

dkijania:
Re-audit of the SDKs (2026-09-21) independently flagged this as an upstream defect: `src/config.ts:22` still lists four of the five root fields, so `ENABLED_QUERIES=verificationKeyUpdates` aborts startup. I re-checked `main` today — `KNOWN_QUERIES = ['events', 'actions', 'networkState', 'blocks']` — and this PR fixes it correctly.

Status: **all 8 checks green**, `mergeStateStatus` is `BLOCKED` solely on `REVIEW_REQUIRED`. It has been in that state since before 2026-09-20.

Flagging rather than touching it: it needs a reviewer, not changes.

